PR #5, opened 2019-10-15
fix: don't allow mutating original config object by the caller

rchl:
`loadConfig` with `merge=false` was using spread operator to
clone  properties from fixture's configuration. Since that is
a shallow copy, child objects were returned by reference and
allowed caller to mutate those objects, affecting further
uses of `loadConfig`. Ensure that is not possible by returning
a copy of loaded config.

This was a problem when one test loaded config and passed it
to Nuxt Builder. Nuxt builder then appended generated plugins
to plugins object and those plugins were then used by subsequent
tests, even if it intended to build new, clean fixture.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -1,3 +1,4 @@
+const clonedeep = require('clone-deep')
 const deepmerge = require('deepmerge')
 const request = require('request-promise-native')
 const getPort = require('get-port')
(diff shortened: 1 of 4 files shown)


rchl:
@ricardogobbosouza thank you. Are you also able to create new release?

PR #14, opened 2020-03-03
Add Types for typescript support

mentAl-maZe:
I wrote some shims in order to be able to use this module with TS and thought it might be nice to have native support for that.

I also took the liberty to update dependencies to use `request-promise` in lieu of the deprecated `request-promise-native`, and to remove the unused dependency `request`.


moleCuleFFF:
+1 Need this :D

PR #22, opened 2020-07-01
fix(types): correct types of get method options

rchl:
"Options" type was not correct since it expands to "request.UriOptions & RequestPromiseOptions"
which means that it requires "uri" or "url" property while this "get" API requires "uri" separately.
Use just "RequestPromiseOptions" for the options.


ricardogobbosouza:
Thanks again @rchl

PR #23, opened 2020-07-21
refactor: update module

ricardogobbosouza:


package.json
```diff
@@ -7,16 +7,15 @@
...
-    "types"
+    "dist"
   ],
-  "main": "lib/index.js",
-  "types": "types/index.d.ts",
   "scripts": {
     "build": "tsc",
-    "lint": "eslint --ext .js,.vue .",
-    "release": "yarn test && standard-version && git push --follow-tags && npm publish",
+    "lint": "eslint --ext .ts .",
```
pi0, line 17: We would also need `.js,.vue` for fixtures

package.json
```diff
@@ -7,16 +7,15 @@
...
+    "release": "yarn test && yarn build && standard-version && git push --follow-tags && npm publish",
```
pi0, line 18: Neat: Probably not needed for us but doing `build` after version, ensures if bundling version, we use updated one

src/nuxt.ts
```diff
@@ -31,8 +30,9 @@ export async function loadFixture() {
   }
 }
 
-export async function loadNuxtPackage(name: 'nuxt') {
+export async function loadNuxtPackage () {
```
pi0, line 33: It is a general utility with type hint :) But change is ok :+1:

package.json
ricardogobbosouza: Fixed

PR #31, opened 2020-07-30
test: add tests

ricardogobbosouza:


src/types.ts
```diff
@@ -8,7 +8,7 @@ export interface NuxtTestContext {
   rootDir: string
   config: NuxtConfig
   nuxt: {
-    config: NuxtConfig
+    options: NuxtConfig
```
pi0, line 11: Nice find! Maybe we should type it to `NuxtOptions` too?

test/generate.test.ts
```diff
@@ -0,0 +1,19 @@
+import { readFileSync } from 'fs'
+import { resolve } from 'path'
+import { setupTest, NuxtTestContext } from '../src'
+
+describe('generate', () => {
+  const ctx: NuxtTestContext = setupTest({
```
pi0, line 6: We may also destructure `{ nuxt }` for tests. Will also remove requirement of importing `NuxtTestContext` type

ricardogobbosouza: For some reason, when it desestruturar the test fails

src/types.ts
ricardogobbosouza: Sure

ricardogobbosouza: Found, thanks to tests :smile:

test/generate.test.ts
pi0: What's the bug? 😬

package.json
```diff
@@ -34,8 +34,10 @@
     "codecov": "latest",
     "eslint": "latest",
     "husky": "latest",
+    "core-js": "2",
```
pi0, line 37: nuxt 2.14 should go fine with 3 too

PR #42, opened 2020-08-05
feat: ensure puppeteer is installed

### LINKED ISSUE #34: [v2] auto install puppeteer
Currently if not having dependency installed will throw an error. We may automate it using [lmify](https://github.com/nuxt-contrib/lmify) or at least throw error with instructions

ricardogobbosouza:
Resolve #34

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -5,6 +5,17 @@ export async function createBrowser () {
   const ctx = getContext()
   const tib = await import('tib')
 
+  if (ctx.browserString === 'puppeteer') {
+    try {
+      require.resolve(ctx.browserString)
+    } catch {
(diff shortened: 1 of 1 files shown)


src/browser.ts
```diff
@@ -5,6 +5,17 @@ export async function createBrowser () {
   const ctx = getContext()
   const tib = await import('tib')
 
+  if (ctx.browserString === 'puppeteer') {
+    try {
+      require.resolve(ctx.browserString)
```
pi0, line 10: Is there always a 1-1 map from browserString to npm package? /cc @pimlie

pi0: NVM 🙈

PR #43, opened 2020-08-05
chore: rename jest matchers

### LINKED ISSUE #37: [v2] rename toNuxtPluginAdded
as suggested by @rchl

ricardogobbosouza:
Resolve #37

### DIFF
--- a/src/jest.matchers.ts
+++ b/src/jest.matchers.ts
@@ -3,23 +3,23 @@ import { NuxtTestContext } from './types'
 declare global {
   namespace jest {
     interface Matchers<R> {
-      toNuxtPluginAdded(plugin: any): CustomMatcherResult
-      toNuxtLayoutAdded(layout: any, name?: string): CustomMatcherResult
-      toNuxtErrorLayoutAdded(errorLayout: any): CustomMatcherResult
-      toNuxtServerMiddlewareAdded(middleware: any): CustomMatcherResult
-      toNuxtRequireModule(moduleOpts: any): CustomMatcherResult
+      toHaveCalledNuxtAddPlugin(plugin: any): CustomMatcherResult
+      toHaveCalledNuxtAddLayout(layout: any, name?: string): CustomMatcherResult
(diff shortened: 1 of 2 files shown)


src/jest.matchers.ts
```diff
@@ -3,23 +3,23 @@ import { NuxtTestContext } from './types'
 declare global {
   namespace jest {
     interface Matchers<R> {
-      toNuxtPluginAdded(plugin: any): CustomMatcherResult
-      toNuxtLayoutAdded(layout: any, name?: string): CustomMatcherResult
-      toNuxtErrorLayoutAdded(errorLayout: any): CustomMatcherResult
-      toNuxtServerMiddlewareAdded(middleware: any): CustomMatcherResult
-      toNuxtRequireModule(moduleOpts: any): CustomMatcherResult
+      toHaveCalledNuxtAddPlugin(plugin: any): CustomMatcherResult
```
pi0, line 6: /cc @danielroe @manniL if you approve it is grammatically correct.

danielroe: This is much better than previous. But what about either of:
- `toHaveAddedNuxtPlugin` <- feels most idiomatic, but I realise the matcher is only testing for the `addPlugin` method to be called
- `toHaveCalledAddNuxtPlugin`

rchl: I might be biassed since I've came up with those suggestions but IMO it's better to match exactly the method call names that it's checking. So I would prefer to have `AddPlugin` there literally rather than inserting `Nuxt` in between.

Another benefit of the current change is that when user types `toHaveCalledNuxt`, he will have all methods suggested in the editor while moving the `Nuxt` part further will make that work less optimally.

(We are not writing a book here so it doesn't have to read smoothly. It's more important that it's accurate and have symmetry with the code it's checking. ;))

manniL: I personally agree with @rchl and think `toHaveCalledNuxt` is a decent way to prefix the matchers.

Issue #50, opened 2020-08-08
feat: run parallel tests with different test contexts


danielroe:
I'd love to be able to initialise `setupTest` several times in the same file or even across files in the test suite.

<details>
<summary>E.g. do something like this</summary>

```js
import { setupTest, createPage, cleanup } from '@nuxtjs/module-test-utils'

describe('module with default options', () => {
  setupTest({
    __dirname,
    browser: true,
    fixture: '../example',
  })

  test('My test', async () => {
    // ...
  })
})

describe('module with other options', () => {
  setupTest({
    __dirname,
    browser: true,
    fixture: '../example',
    config: differentOptions,
  })

  test('My test', async () => {
    // ...
  })
})
```

</details>

If I do this, I get non-determinative behaviour - or errors - even if the test is setup only once in each test spec file. (That is, options *leak* from one test context into another.) Or (if I pass different config objects in different files), I get:
```bash
 TypeError: Cannot read property 'moduleContainer' of undefined

   at node_modules/@nuxtjs/module-test-utils/dist/index.js:286:29
   at fulfilled (node_modules/tslib/tslib.js:112:62)
```

pi0:
According to [jest docs](https://jestjs.io/docs/en/setup-teardown#order-of-execution-of-describe-and-test-blocks) tests in a single file are always executed serially but issue is that inside `setupTest`, we use `beforeAll` which jest executes all of them first (still serially but reordered regardless of describe scope)

Something strange is that I tried changing `beforeAll(() => {})` to `test('build', () => {})` inside `setupTest` but jest still behaves it similar to beforeAll (probably I missed something to try the fix). Only possible way to fix was:

```js
  test('setup', async () => {
    await setupTest({
      __dirname,
      browser: true,
      fixture: '../example',
    })
  }, 60000)
```

danielroe:
> **Once the describe blocks are complete**, by default Jest runs all the tests serially in the order they were encountered in the collection phase, waiting for each to finish and be tidied up before moving on.

That sounds to me like as though where multiple describe blocks in the same file set up different contexts, there will be a conflict. In fact, I've tested further and there is a conflict even when multiple files set up different contexts. The tests at least *run* but one context overwrites the other.

I'm currently wondering if [a custom environment](https://jestjs.io/docs/en/next/configuration#testenvironment-string) might solve the problem and actually add some features/DX.

danielroe:
[...]

Issue #38, opened 2020-08-05
[v2] rewrite proxy for `setupTest`


ricardogobbosouza:
```js
describe('module', () => {
  const { nuxt } = setupTest({
    __dirname,
    fixture: 'example'
  })

  test('render',() => {
    // nuxt undefined
  })
})
```

pi0:
This won't be necessary anymore with #51. One can use `getNuxt()`

PR #51, opened 2020-08-09
feat: keep internal context

pi0:
## Refactors / Features

- Don't return `ctx` from `setupTest`: This ensures no context-sharing will happen because of wrong usage also fixing destructure issue (#38) by introducing new `getNuxt()` utility (or for advanced use cases one can directly use `getContext()`)
- Refactor jest matchers into contextless assertions to avoid passing ctx (generic `expectModuleToBeCalledWith` util) 
- Use `afterEach` / `beforeEach` to ensure all scoped tests get their own context and tests don't have side-effect to each other

## Fixes

- Because jest calls all `beforeAll` hooks in single test before all regardless of `describe` scope, switching from `beforeAll` to `test` which guarantees sequential order for setup stage fixing (#50)


src/assertions.ts
```diff
@@ -0,0 +1,15 @@
+import { getNuxt } from './nuxt'
+
+type ModuleContainerMethod = 'addPlugin' | 'addLayout' | 'addErrorLayout' | 'addServerMiddleware' | 'requireModule'
+
+export function expectModuleToBeCalledWith (method: ModuleContainerMethod, ...args) {
+  const _args = []
```
pi0, line 6: My bad. Reverted by  d1c1c7192c8ba2ffd3609f42ac9f431d6ded474f

PR #53, opened 2020-08-10
fix: use random `buildDir` to avoid conflicts

### LINKED ISSUE #52: fix: build conflict between multiple tests
Originally reported by #50, when having multiple `.test` files, jest runs them in parallel. If two depending on a single fixture and need build, this causes issues. Some options are:

- Using run-in-band to disable parallel run
- Using a lock-file so one test awaits on others during build
- Using random build ids for `buildDir`
- Using mfs (#49) for building in mem (considering all internals can support)

ricardogobbosouza:
Resolve #52

### DIFF
--- a/src/nuxt.ts
+++ b/src/nuxt.ts
@@ -24,7 +24,7 @@ export async function loadFixture () {
   }
 
   if (!ctx.config.buildDir) {
-    ctx.config.buildDir = resolve(ctx.rootDir, '.nuxt', ctx.id.toString())
+    ctx.config.buildDir = resolve(ctx.rootDir, '.nuxt', ctx._id.toString())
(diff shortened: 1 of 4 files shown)


src/nuxt.ts
```diff
@@ -22,6 +22,10 @@ export async function loadFixture () {
   if (!ctx.config.rootDir) {
     ctx.config.rootDir = ctx.rootDir
   }
+
+  if (!ctx.config.buildDir) {
+    ctx.config.buildDir = resolve(ctx.rootDir, '.nuxt', ctx._id.toString())
```
pi0, line 27: This id is not unique across multiple test files unless manually set...

PR #64, opened 2020-09-30
fix(generate): start nuxt server before generate

farnabaz:
There is also another way to start a server, by moving
```js
    if (ctx.options.server) {
      await listen()
    }
```
before
```js
   if (ctx.options.generate) {
      await generate()
    }
```
in the [setup.ts](https://github.com/nuxt-community/module-test-utils/blob/master/src/setup.ts), but it has two downsides:
1. It needs to set `server: true` by user.
2. It does not work without `setupTest` functions

### DIFF
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -9,5 +9,7 @@ export async function generate () {
   const builder = new Builder(nuxt)
   const generator = new Generator(nuxt, builder)
 
-  await generator.generate(options.generate)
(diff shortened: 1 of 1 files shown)


pi0:
Thanks! @farnabaz Would you please fix lint issue?

pi0:
>     if (ctx.options.server) {
      await listen()
    }

Using alternative way would be better since we can use cleanup hook and only listen to one port

PR #79, opened 2021-01-14
docs(example): add missing `expectModuleToBeCalledWith` in example

Geminii:
Into the nuxt module example, it miss the import of `expectModuleToBeCalledWith`


pi0:
THanks!

PR #80, opened 2021-01-16
docs: fix nuxt/test-utils preview on README

Geminii:
Currently preview image is broken because of wrong extension


pi0:
Thanks!

PR #81, opened 2021-01-20
docs: migration to docus

Geminii:
Migration to new beautiful docus documentation ❤️

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -6,9 +6,8 @@
     "start": "nuxt start"
   },
   "devDependencies": {
+    "docus": "^0.0.4",
     "nuxt-edge": "latest"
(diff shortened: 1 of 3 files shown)


docs/package.json
```diff
@@ -6,7 +6,9 @@
     "start": "nuxt start"
   },
   "devDependencies": {
-    "@nuxt/content-theme-docs": "latest",
     "nuxt-edge": "latest"
+  },
+  "dependencies": {
```
pi0, line 11: Shall we keep both in same section? (`dependencies` or `devDependnecies`)

Geminii: Yes of course, you are right 👍

PR #84, opened 2021-02-08
fix(server): try to use `config.server.port`

vanhoofmaarten:
Respect `server.port` configuration, set in `setupTest` or `nuxt.config.js`

eg. 
```js
import { setupTest  } from '@nuxt/test-utils'

const port = 3456

setupTest({
  browser: true,
  config: {
    server: {
      port
    }
  }
})
```

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,7 +5,7 @@ import { getContext } from './context'
 export async function listen () {
   const ctx = getContext()
   const { server } = ctx.options.config
-  const port = server?.port ? Number(server.port) : await getPort()
+  const port = await getPort({ port: server?.port?.port })
 
   ctx.url = `http://localhost:${port}`
 
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,9 +5,11 @@ import { getContext } from './context'
 export async function listen () {
   const ctx = getContext()
   const { server } = ctx.options.config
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -4,9 +4,10 @@ import { getContext } from './context'
 
 export async function listen () {
   const ctx = getContext()
-  const port = await getPort()
+  const { server } = ctx.options.config
+  const port = server?.port ? Number(server.port) : await getPort()
```
pi0, line 8: We can pass server.port as [default port](https://www.npmjs.com/package/get-port) otherwise an error happens if port is already used

```suggestion
  const port = await getPort({ port: server?.port?.port })
```

vanhoofmaarten:
Already thanks for the review, @pi0. I've gone a refined a little further, catching undefined values coming from the config that might result in an error from `getPort`.

eg. When passing `undefined` the `Number` method:
`RangeError: options.port should be >= 0 and < 65536. Received NaN.RangeError [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536. Received NaN.`

pi0 (approved):

pi0:
LGTM. Thanks!

vanhoofmaarten:
My pleasure. Thanks for the quick release!

PR #87, opened 2021-02-20
feat(nuxt): allow function and promise config

vanhoofmaarten:
Allow `loadFixture`to accept Nuxt configs declared as functions or promises.

Taken from https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/load.js#L61-L71

### DIFF
--- a/src/nuxt.ts
+++ b/src/nuxt.ts
@@ -9,13 +9,38 @@ export async function loadNuxt () {
   ctx.nuxt = new Nuxt(ctx.options.config)
 }
 
+function loadConfig () {
+  const { options } = getContext()
+  return (
+    import(resolve(options.rootDir, options.configFile))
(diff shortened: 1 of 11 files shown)


ricardogobbosouza (approved):

pi0:
Thanks for PR @vanhoofmaarten. I've changed package to directly use `loadNuxtConfig` exposes by nuxt from 2.12+. This way we have most consistent behavior with core.

However it is most likely in next versions (and when migrating to kit) this method will be deprecated since we cannot statically analyze config anymore. So i would suggest migrating to alternative approach by using local async module or hooks to do async setup.

vanhoofmaarten:
Hi @pi0. Good thinking! Thanks for the time and effort.

PR #91, opened 2021-03-05
feat: mockConsola

### LINKED ISSUE #90: Helper to mock consola
We often need to do something like this:

```js
const mockReporter = {
  error: jest.fn(),
  info: jest.fn(),
  success: jest.fn()
}

jest.mock('consola', () => ({
  info: jest.fn(),
  success: jest.fn(),
  debug: jest.fn(),
  withTag: jest.fn().mockImplementation(() => mockReporter)
}))
```

A helper for that would be nice :)

ricardogobbosouza:
Resolve #90


src/mocks.ts
```diff
@@ -0,0 +1,19 @@
+import consola from 'consola'
+
+export function mockConsola () {
+  const mock = consola
```
pi0, line 4: Is there any reason to rename consola to mock?

test/fixtures/consola-with-tag/nuxt.config.js
```diff
@@ -0,0 +1,11 @@
+import consola from 'consola'
+
+module.exports = {
+  hooks: {
+    build: {
+      done () {
+        consola.withTag('build').error('bar')
```
pi0, line 7: We can move both to basic fixture to avoid building/setupTest overhead

src/mocks.ts
ricardogobbosouza: To completely mock, it does not work using the global object `consola`

pi0: They are same reference....

test/fixtures/consola-with-tag/nuxt.config.js
ricardogobbosouza: Updated

src/mocks.ts
ricardogobbosouza: You're right :satisfied:

src/mocks.ts
```diff
@@ -0,0 +1,13 @@
+import consola from 'consola'
+
+export function mockConsola (): typeof consola {
+  const mock = {}
+
+  consola.mockTypes((type) => {
+    mock[type] = mock[type] || jest.fn()
```
pi0, line 7: /cc @ricardogobbosouza. Consola does inherit mocked logs but here we should remember first mocked version otherwise after creating tag, a new mock (jest.fn()) will be created

pi0 (approved):
Merging for release. PR for docs more than welcome and thanks!

Issue #98, opened 2021-03-09
page https://test-utils.nuxtjs.org/options not found


deni5n:
link "See module options" in https://github.com/nuxt/test-utils not found
link "Setup" in https://test-utils.nuxtjs.org/api-reference/setup not worked

ricardogobbosouza:
Resolved #103

PR #103, opened 2021-03-30
docs: fix broken links

### LINKED ISSUE #102: Bug: misconfigured next chapter link
In [setup page](https://test-utils.nuxtjs.org/api-reference/setup/) next chapter link is references the current page instead of 'Testing an app' page. 
image

mrazauskas:
Fixes #102.

Two files were named `setup.md` and it seems like this was confusing Docus. I renamed one of the files and next chapter link seems to be working as expected.


mrazauskas:
Fixes #98.

And another quick fix for broken link in readme file. I see the intention to rewrite docs. That’s good idea. Most probably everything will change, but perhaps it still makes sense to fix this broken link.

mrazauskas:
Thanks. I’m glad it helps.

Unfortunately it seems like nothing changed on the website. Can it be that the build failed?

At least on my machine docs do not build. `docus@^0.0.9` and `nuxt-edge@latest` are the only dependancies. I was trying to  upgrade to `docus@^0.1.5` and it solved the problem. It would be enough to merge #101.

If there is a reason to avoid #101, alternative solution could be to install `nuxt@^2.15.3` instead of `nuxt-edge`. Also works. I can send a PR.

PR #106, opened 2021-03-31
fix(types): add more type definitions to nuxt interface

### LINKED ISSUE #104: renderAndGetWindow type not found
https://github.com/nuxt-community/laravel-echo-module/blob/main/test/basic.test.ts#L12

ricardogobbosouza:
Some definitions were not in context

Resolve #104


pi0 (approved):
Thanks! We should probably backport it to `@nuxt/types` for 2.16

PR #107, opened 2021-03-31
feat: resolve rootDir

### LINKED ISSUE #96: Detect rootDir if not defined
In the beginning the objective of this package was to test modules and with that its initial context was defined for that.

To test applications built with `nuxt` we usually have to use a root folder in the config [`fixture`](https://github.com/nuxt/test-utils/blob/main/src/context.ts#L59).

We may be able to detect this automatically:
- When using modules, use a folder `fixture`
- When applying, use the project root

ricardogobbosouza:
Resolves #96, #76

### DIFF
--- a/src/nuxt.ts
+++ b/src/nuxt.ts
@@ -9,6 +9,14 @@ export async function loadNuxt () {
   ctx.nuxt = new Nuxt(ctx.options.config)
 }
 
+const isNuxtApp = (dir: string) => {
+  return existsSync(dir) && (
+    existsSync(resolve(dir, 'pages')) ||
+    existsSync(resolve(dir, 'nuxt.config.js')) ||
+    existsSync(resolve(dir, 'nuxt.config.ts'))
+  )
+}
+
 const resolveRootDir = () => {
   const { options } = getContext()
 
@@ -17,14 +25,6 @@ const resolveRootDir = () => {
     resolve(options.testDir, options.fixture),
     process.cwd()
   ]
-
-  const isNuxtApp = (dir: string) => {
-    return existsSync(dir) && (
-      existsSync(resolve(dir, 'pages')) ||
-      existsSync(resolve(dir, 'nuxt.config.js')) ||
-      existsSync(resolve(dir, 'nuxt.config.ts'))
-    )
-  }
 
   for (const dir of dirs) {
     if (dir && isNuxtApp(dir)) {
--- a/src/nuxt.ts
+++ b/src/nuxt.ts
@@ -11,7 +11,12 @@ export async function loadNuxt () {
 
 const resolveRootDir = () => {
   const { options } = getContext()
-  const dirs = [options.rootDir, resolve(options.testDir, options.fixture), process.cwd()]
+
+  const dirs = [
+    options.rootDir,
+    resolve(options.testDir, options.fixture),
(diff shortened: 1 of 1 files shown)


mrazauskas:
For apps it might be better to set `options.fixture = process.cwd()` instead of `options.rootDir = process.cwd()`. The app is the fixture in this case. So obvious, but took me few days to understand.

If the app is at `/the-app` folder and `options.fixture = process.cwd()` is set `loadNuxtConfig` is called with:

testDir: `/the-app/test`
rootDir: `/the-app`
buildDir: `/the-app/.nuxt/<randomID>`

Seems perfect for app testing or I missed something? If I get it right `loadFixture` function simply overwrites anything what was set in `options.rootDir`. To avoid calling `loadFixture` one could pass `options.fixture = false`.

Another thing – how do we know if an app or a module is under test? It might be I found a solution. Playing with working prototype right now. Will send in a PR in few days.

pi0 (changes requested):
- Value of `rootDir`, can be determined by searching in an array of `[options.rootDir, resolve(options.testDir, options.fixture), process.cwd()]`
  - We have a match if app dir exists and app/pages or app/nuxt.config exists
  - If non of array items do match, we should throw an error asking to explicitly set `options.rootDir` pointing to a valid nuxt app

mrazauskas:
What if `dir/pages` is found, but these are just some pages, but not Nuxt routes? Don’t get me wrong, I agree that it would be very smart to do this check. At the same time `dir/pages` is so generic.

Might work, of course. But in some case it will fail. Perhaps it would be helpful to `console.info` why the algorithm made one or another assumption. One could see what the app was trying to do and then it is easier to debug.

src/nuxt.ts
```diff
@@ -9,16 +9,32 @@ export async function loadNuxt () {
   ctx.nuxt = new Nuxt(ctx.options.config)
 }
 
-export async function loadFixture () {
+const resolveRootDir = () => {
   const { options } = getContext()
+  const dirs = [options.rootDir, resolve(options.testDir, options.fixture), process.cwd()]
+
+  const isNuxtApp = (dir: string) => {
```
pi0, line 21: It would be nice movie it to a top level function

src/nuxt.ts
```diff
@@ -9,16 +9,32 @@ export async function loadNuxt () {
...
+      existsSync(resolve(dir, 'nuxt.config.ts'))
+    )
+  }
 
-  options.rootDir = resolve(options.testDir, options.fixture)
-
-  if (!existsSync(options.rootDir)) {
-    // define root app if not exists test/fixture
-    options.rootDir = resolve(process.cwd())
+  for (const dir in dirs) {
```
pi0: I think should be for of? :)

src/nuxt.ts
ricardogobbosouza: I thought about it, and also about exporting it

ricardogobbosouza: @pi0 I don't know if it makes a lot of sense to export them, does it?

pi0: Thanks. As test-utils probably not but we may add it to nuxt/kit (i will invite you to repo soon btw)

pi0 (approved):

PR #108, opened 2021-04-09
refactor: Node’s native functionality used to get random port

mrazauskas:
*Typings.* `nuxt.listen()` handles numbers and strings as `port` property (see test). For example, a `string` might be useful if config is read using `process.env`.

*`URL` class.* I had to add it, because `nuxt.listen()` returns url with trailing slash included. That’s why `ctx.url + '/'` did not work. In general URL makes usage more flexible for the user (see test).

*`get-port`.* Was factoring out. I was looking at `get-port`’s code and could not fully understand how it works. Went to check Node’s docs on `server.listen()`:

> If port is omitted or is 0, the operating system will assign an arbitrary unused port...

It means that by calling `nuxt.listen(0)` we get random port with having a dependancy. Factoring out also effectively mitigates `get-port`’s [tiny chance of a race condition](https://github.com/sindresorhus/get-port#beware).

*Benchmark.* Checked speed, but I did not find any before / after difference.

*Tests.* I was trying to see if there is any need to validate the `port` value before passing it to `nuxt.listen()`. Only `number` and `string` are allowed in typings, so I was focusing only on these two (not arrays, objects, etc). Seems like `nuxt`, `connect` and `node` handle everything properly and throw errors. There is no need to validated the value. Tests were left in place as a proof. Good to have them for the future as well.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,17 +1,14 @@
-import getPort from 'get-port'
 import got, { OptionsOfUnknownResponseBody } from 'got'
 import { getContext } from './context'
 
 export async function listen () {
   const ctx = getContext()
   const { server } = ctx.options.config
-  const port = await getPort({
-    ...(server?.port && { port: Number(server?.port) })
-  })
 
-  ctx.url = 'http://localhost:' + port
+  const port = server?.port || 0
+  const { url } = await ctx.nuxt.listen(port)
 
-  await ctx.nuxt.listen(port)
+  ctx.url = url
 }
 
 export function get (path: string, options?: OptionsOfUnknownResponseBody) {
@@ -25,5 +22,7 @@ export function url (path: string) {
     throw new Error('server is not enabled')
   }
 
-  return ctx.url + path
+  const { href } = new URL(path, ctx.url)
+
+  return href
 }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "@babel/preset-typescript": "^7.13.0",
     "consola": "^2.15.3",
     "defu": "^3.2.2",
-    "get-port": "^5.1.1",
     "got": "^11.8.2"
   },
   "devDependencies": {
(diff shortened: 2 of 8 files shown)


src/server.ts
```diff
@@ -1,17 +1,14 @@
...
 export async function listen () {
   const ctx = getContext()
   const { server } = ctx.options.config
-  const port = await getPort({
-    ...(server?.port && { port: Number(server?.port) })
-  })
 
-  ctx.url = 'http://localhost:' + port
+  const port = server?.port || 0
+  const { url } = await ctx.nuxt.listen(port)
```
pi0, line 9: It would be nice if you can replace it with [listhen](https://github.com/unjs/listhen) to get url (`listen(ctx.nuxt.app)`). Since this listen function is removed in nuxt3. Also if need to normalize url, can use [ufo](https://github.com/unjs/ufo).

pi0: Also `get` implementation should be updated with [this](https://github.com/unjs/ohmyfetch). These are standard utilities we use across nuxt repos. I didn't had a chance yet to update them.

mrazauskas: Thanks. This is very useful. I will fix it.

mrazauskas: Could you give hint how nuxt.app() should be typed?

pi0: Here: https://github.com/nuxt/test-utils/blob/a5c7b753412c4600a2ebe4ca831f0f52f68142b1/src/context.ts#L37

we need to add it like `app: RequestListener` (`import type { RequestListener } from 'http'`)

mrazauskas: Got it. All works here. Fixing the rest

mrazauskas:
To move on I would like to:

- [ ] add one feature to `get-port-please`
- [ ] make one optimisation inside `listhen`
- [ ] find out why `got` is not working with `listhen` (did not look at it carefully and, of course, will migrate to `ohmyfetch`, but would be good to know what didn’t work).

Working on this. Really fun to play with these small libraries.

pi0:
get-port update (2.2.0) released

mrazauskas:
Found better solution.

PR #112, opened 2021-04-26
docs: clarification between nuxt and components testing

### LINKED ISSUE #111: @nuxt/test-utils vs @vue/test-utils?
It would be really good to know why using `@nuxt/test-utils` is more beneficial than using `@vue/test-utils` in a Nuxt project. Despite the obvious point being this package is made specifically for Nuxt, `@vue/test-utils` still works perfectly well in Nuxt projects.

Could we get some comparison docs (a table maybe) of what `@vue/test-utils` can't do for a Nuxt project, but `@nuxt/test-utils` can?

MentalGear:
Based upon:
https://github.com/nuxt/test-utils/issues/111
@mrazauskas 

Resolves #111

### DIFF
--- a/docs/content/en/index.md
+++ b/docs/content/en/index.md
@@ -20,7 +20,8 @@ Now it's easy to create tests for projects and modules using Nuxt.
 
 <list :items="items"></list>
 
-## Nuxt Framework vs Vue Component Testing
-@nuxt/test-utils will build an app (or generate a website), launch a server and will provide a browser (all are optional). For example, @nuxt/test-utils might help to check if Nuxt routes, middleware, modules, etc work and if pages are rendered as expected (e.g. collection of components). Focus is on routes, browsing and Nuxt instance, but not Vue components.
+## Why `@nuxt/test-utils`?
 
-Use [@vue/test-utils](https://vue-test-utils.vuejs.org/) to test single Vue components. (vue/test-utils does not build Nuxt apps and knows nothing about Nuxt instances)
+This library allows you to test your Nuxt app or module. It includes unit and end-to-end test helpers, but is not intended as a substitute for component test libraries like `@vue/test-utils`.
+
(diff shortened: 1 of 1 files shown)


docs/content/en/index.md
```diff
@@ -19,3 +19,8 @@ Now it's easy to create tests for projects and modules using Nuxt.
 ## Key features
 
 <list :items="items"></list>
+
+## Nuxt Framework vs Vue Component Testing
+@nuxt/test-utils will build an app (or generate a website), launch a server and will provide a browser (all are optional). For example, @nuxt/test-utils might help to check if Nuxt routes, middleware, modules, etc work and if pages are rendered as expected (e.g. collection of components). Focus is on routes, browsing and Nuxt instance, but not Vue components.
+
+Use [@vue/test-utils](https://vue-test-utils.vuejs.org/) to test single Vue components. (vue/test-utils does not build Nuxt apps and knows nothing about Nuxt instances)
```
danielroe, line 26:
```suggestion
## Why `@nuxt/test-utils`?

This library allows you to test your Nuxt app or module. It includes unit and end-to-end test helpers, but is not intended as a substitute for component test libraries like `@vue/test-utils`.

With this library you can interact with a Nuxt instance - whether that is verifying a plugin or module is behaving as expected, getting a fully-rendered server response or conducting client-side browser tests using playwright.
```

MentalGear: Thanks, this makes things much clearer for newcomers, especially adding "e2e" while now mentioning @vue/test-utils for component testing.

Just one question: You write "it includes unit ... test helpers". Can you specify what unit helpers are, in context of nuxt-test-utils?
For example, atm I use ts-jest to test my functions that are separated into files. How would this package help me do unit testing?

ricardogobbosouza (approved):

PR #120, opened 2021-05-07
fix: add missing method types for moduleContainer

### LINKED ISSUE #119: Module testing page is missing the "addModule" option
in [module testing docs](https://test-utils.nuxtjs.org/api-reference/module-testing) page It only shows the following options:
- `addPlugin`
- `addLayout` 
- `addErrorLayout`
- `addServerMiddleware` 
- `requireModule`

But the `addModule` option  is not documented although it can be used!

achimoraites:
Included the `addModule` option in the [module testing page](https://test-utils.nuxtjs.org/api-reference/module-testing)

fixes #119

### DIFF
--- a/docs/content/en/api-reference/module-testing.md
+++ b/docs/content/en/api-reference/module-testing.md
@@ -16,7 +16,7 @@ When you write tests for a Nuxt module, you normally expect the module to be ins
 * `method`
   * Type: `String`
   * `required`
-  * Available values are `addModule`, `addPlugin`, `addLayout`, `addErrorLayout`, `addServerMiddleware` and `requireModule`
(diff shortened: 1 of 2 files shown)


mrazauskas:
Would be good to fix [typings](https://github.com/nuxt/test-utils/blob/e3ad52ef091a8b7a541b2add0fda4894626e7eb8/src/assertions.ts#L5) as well.

achimoraites:
Thanks a lot @mrazauskas !
I will fix them as well :)

ricardogobbosouza (approved):

pi0:
Also missing: `addTemplate`, `extendBuild`, `extendBuild`, `addModule` (reference: https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/module.js)

achimoraites:
Thanks @pi0 !
Will include them as well!

pi0 (approved):

PR #122, opened 2021-05-14
chore(examples): add usage examples

### LINKED ISSUE #116: Conflict between @nuxt/test-utils jest preset and @vue/test-utils
Hi,

I was trying to integrate @nuxt/test-utils in my Nuxt application in order to test pages.
I already have a bunch of tests written for the components that use @vue/test-utils.

After following the documentation and adding the `@nuxt/test-utils` preset to my jest configuration, all my components tests started failing with the same error:

```shell
[vue-test-utils]: window is undefined, vue-test-utils needs to be run in a browser environment. 
    You can run the tests in node using jsdom 
    See https://vue-test-utils.vuejs.org/guides/#browser-environment for more details.
```

Any idea how to solve this and be able to use both tools within the same project ?

Thanks!

mrazauskas:
Resolves #116, #118

Perhaps it would be helpful to add few basic examples to the repo? Might be useful to understand how everything works.

~~I would somehow link them in the docs too. Quit many files being added already, so I left the docs for another PR. Better to concentrate on code first:~~

~~- is Babel anyhow useful in TS example?~~
~~- `create-nuxt-app` placeholder files and dirs. Would it make sense to remove these? How to make all look more clear?~~
~~- `babel-jest`, `ts-jest`, `ts-node`, `vue-jest` are also sort-of testing helpers, but not mentioned in READMEs. I was trying to keep the READMEs short.~~

Made them as lean as I could. Should be easy to get the idea.

What you think?


examples/app-setup-javascript/jest.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
pi0, line 1: Can we please use preset from test-utils? [https://test-utils.nuxtjs.org/setup#configure-jest]. This way we don't need `.babelrc`.

pi0: PS: We may simplify by merging e2e to jest.config and only include overrides. Since if the purpose of example is app testing, module testing is not needed anymore (at least for a starter example)

examples/app-setup-javascript/jsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
pi0, line 1: ~> preset

mrazauskas:
@pio Just tried to remove `.babelrc`, but Vue Test Utils needs. [Docs](https://vue-test-utils.vuejs.org/installation/#using-vue-test-utils-with-jest-recommended)

I added preset in both examples, just to follow Nuxt Test Utils dos. You are right.

Would it be fine to update the preset and docs synchronising them with Vue Test Utils? Namely: `.babelrc` for JS and `ts-jest` for TS. JS setup does not need any dependency to compile TS and vice versa. Also most probably the dependancies are already added (for example by Nuxt TypeScript module). If this sounds good, I can send another PR.

pi0:
Thanks for the changes @mrazauskas and for explaining. You are right makes sense for non-ts files. (there is a second way to let ts-jest transpile `.js` files as well (see `https://kulshekhar.github.io/ts-jest/docs/getting-started/presets) 

Regarding performance and simplicity, I was thinking to try esbuild (maybe [esbuild-jest](https://github.com/aelbore/esbuild-jest)) this way we don't need `babelrc`, extra setup for typescript and probably much better overall performance.

mrazauskas:
`esbuild-jest` looks interesting. In the other hand, I found `ts-jest`’s typecheks useful if test includes more setup code. It is fine to run `tsc` before building library, but to run it each time before running a test is clumsy or not possible in watch mode.

My idea was not to include any of them inside preset. Just to mention possible options in docs and users can choose. Might be they already have a preference. So not including makes the whole setup more flexible and free of unnecessary dependancies.

There are just three things to mention:
- setup for JS
- conservative TS setup with type checks
- performant TS setup without type checks

Just to give simple and clear guidelines. And I would leave Jest preset only for Nuxt specific tasks:
- Node test environment (or even better: Nuxt test environment)
- custom matchers
- global setup / teardown

Of course, this is not related with examples and this PR. But these examples helped to think about the wider picture.

mrazauskas:
@pi0 Perhaps it is fine to land this PR. The examples are in sync with docs and that’s all we need for now.

And later I will send in another PR regarding the preset. With more elaborated motivation. So we can tryout and look at real solution.

PR #124, opened 2021-05-16
chore: remove `buildDir` option in favor of `config.buildDir`

mrazauskas:
Value of the `buildDir` option was never read. It was only included in typings and documentation.

I think it is better to remove it and simplify public API . Seems like there was no need for it.

Couple of days ago I was trying to play with it. Config overrides was perfect solution (also works for `generate.dir`).


pi0:
Thanks

PR #125, opened 2021-05-16
perf: if defined, use `rootDir`

mrazauskas:
If `rootDir` is defined, we can simply use that value.

(Also it seems like after #107 was merged, line 51 is unreachable.)

### DIFF
--- a/src/nuxt.ts
+++ b/src/nuxt.ts
@@ -21,7 +21,6 @@ const resolveRootDir = () => {
   const { options } = getContext()
 
   const dirs = [
-    options.rootDir,
     resolve(options.testDir, options.fixture),
     process.cwd()
   ]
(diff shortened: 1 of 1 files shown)


src/nuxt.ts
```diff
@@ -38,22 +37,18 @@ const resolveRootDir = () => {
...
   const { loadNuxtConfig } = await loadNuxtPackage()
   options.config = await loadNuxtConfig({
-    rootDir: options.rootDir,
+    rootDir,
     configFile: options.configFile,
     configOverrides: options.config
   })
 
-  if (!options.config.rootDir) {
-    options.config.rootDir = options.rootDir
```
pi0, line 51: `config.rootDir` is passed to nuxt (core) config. `options` is for test-utils context.

src/nuxt.ts
```diff
@@ -38,22 +37,18 @@ const resolveRootDir = () => {
 export async function loadFixture () {
   const { options } = getContext()
 
-  options.rootDir = resolveRootDir()
+  const rootDir = options.rootDir || resolveRootDir()
```
pi0, line 40: We still should warn users if rootDir is not a valid nuxt app. What's the point of this change?

mrazauskas:
Apologies this was messy PR. Found better solution.

PR #127, opened 2021-05-19
fix: use `join` instead of `resolve` for paths related with `fixture` option

mrazauskas:
## Problem

Documented behaviour:

`testDir`, default value: '~~/test'
`fixture`, "a fixture directory (under `testDir`)", default value: 'fixture'
`rootDir`, default value: \<testDir>/\<fixture>

All works as expected if `setupTest` is called this way:

```js
setupTest({
  fixture: 'some-fixture'
})

rootDir === '<process.cwd()>/test/fixture/some-fixture' // true
```

But it will not work with leading slash:

```js
setupTest({
  fixture: '/some-fixture',
})

rootDir === '/some-fixture' // true
```

`path.resolve` is called inside `resolveRootDir` which returns `rootDir` value.

## Solution

`path.join` instead of `path.resolve` does the right job.

By the way, `expectFileToBeGenerated` and `expectFileNotToBeGenerated` have the same issue. Calling `expectFileToBeGenerated('/index.html')` will not check if the file exists inside `generate.dir`, but will look for it at system root. Again `path.join` solves the problem.

Elsewhere `path.resolve` works as expected and there is no need for any change. In the other hand, `path.join` is enough, `nuxt.ts` file would have one import less, and also `path.join` has shorter name, this makes bundle size smaller. Minor details, but it is fun to keep them in mind.

## Alternative Solution

Update docs with a warning to avoid leading slash or to use absolute paths. Also possible.

### DIFF
--- a/src/assertions.ts
+++ b/src/assertions.ts
@@ -1,5 +1,5 @@
 import { existsSync } from 'fs'
-import { resolve } from 'path'
+import { join } from 'path'
 import { getNuxt } from './nuxt'
 
 type ModuleContainerMethod = 'addTemplate' | 'extendBuild' | 'extendRoutes' | 'addModule' | 'addPlugin' | 'addLayout' | 'addErrorLayout' | 'addServerMiddleware' | 'requireModule'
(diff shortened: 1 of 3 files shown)


ricardogobbosouza (approved):

pi0:
I think it was an expected behavior and should revert this or @ricardogobbosouza. Depending on implicit relative paths is not good idea. We can instead support passing `rootDir` to `setupTest` to disable `fixture` addition.

mrazauskas:
@pi0 You are right. I should improve this PR. Better to leave `join` only in places where first argument is always absolute path. Otherwise `resolve` does better job.

ricardogobbosouza:
@pi0 what's your idea about it ? I did not understand

> We can instead support passing rootDir to setupTest to disable fixture addition.

PR #131, opened 2021-06-30
feat!: use ohmyfetch

ricardogobbosouza:
Use `ohmyfetch` instead of `got`

### Breaking change
The result of `ohmyfetch` is different of `got`

```js
// ohmyfetch
const html = await get('/')
expect(html).toContain('Works!')

// got
const { body } = await get('/')
expect(body).toContain('Works!')
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "consola": "^2.15.3",
     "defu": "^5.0.0",
     "get-port": "^5.1.1",
-    "got": "^11.8.2"
+    "ohmyfetch": "^0.2.0"
   },
   "devDependencies": {
     "@babel/plugin-transform-runtime": "latest",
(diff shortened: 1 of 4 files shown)


pi0 (approved):

mrazauskas:
Perhaps it might be useful to include the [typed response](https://github.com/unjs/ohmyfetch#%EF%B8%8F-type-friendly) feature as well?

```ts
export function get<T> (path: string, options?: FetchOptions) {
   return $fetch<T>(url(path), options)
 }
```

pi0:
Looks nice @ricardogobbosouza. Do you mind also updating docs usage?

ricardogobbosouza:
@pi0 I will do this in another PR, I need to review all documentation

PR #132, opened 2021-06-30
fix: add `transformIgnorePatterns`

### LINKED ISSUE #128: "Cannot use import statement outside a module" while using with nuxt-i18n
Try running `yarn test`.

Error: https://codesandbox.io/s/optimistic-maxwell-l1dmx?file=/tests/test.spec.js

ricardogobbosouza:
Resolve #128

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -12,5 +12,8 @@ module.exports = {
         plugins: ['@babel/plugin-transform-runtime']
       }
     ]
-  }
+  },
+  transformIgnorePatterns: [
+    'node_modules/(?!(nuxt-i18n)/)'
+  ]
 }


jest-preset.js
```diff
@@ -12,5 +12,8 @@ module.exports = {
         plugins: ['@babel/plugin-transform-runtime']
       }
     ]
-  }
+  },
+  transformIgnorePatterns: [
+    'node_modules/(?!(nuxt-i18n)/)'
```
pi0, line 17: nuxt-i18n is not the only module exporting esm syntax....

ricardogobbosouza: No doubt, do we need to make a list of the modules, or do I know another way?

pi0: Actually, I'm thinking to transpile `node_modules` (with auto esm detect system similar to jiti). There is a WIP implementation of esbuild transform that will make public soon :)

ricardogobbosouza: @pi0  Thinking about using
https://www.npmjs.com/package/@swc/jest
or
https://github.com/aelbore/esbuild-jest
?
Any indications?

pi0: I tried both. swc was so old and esbuild-jest is actually using babel!! (source: https://github.com/aelbore/esbuild-jest/blob/master/src/transformer.ts)

Publishing npm package this afternoon. it is inlined in nuxt3 repo.

ricardogobbosouza: We could have [`jiti/jest`](https://github.com/unjs/jiti), no?
So it would be under our supervision.

ricardogobbosouza: With this configuration its works, but increased runtime a little

```js
module.exports = {
  collectCoverage: true,
  testEnvironment: 'node',
  transformIgnorePatterns: [],
  transform: {
    '^.+\\.(t|j)sx?$': [
      'esbuild-jest'
    ]
  }
}
```

another approach would be to ignore only the modules used in nuxt.config

ricardogobbosouza: I don't recommend transpiling all node_modules, often it's just one module or another, it's better to put a warning in the documentation

pi0: Well slowly more and more npm packages are going to use `type: module` so we need to be also futuristic. But also we can use [jiti's approach](https://github.com/unjs/jiti/blob/57f0fbf05d27f98d6c369354b3f47d29f5732782/src/jiti.ts#L175) to smartly transpile when needed :)

Issue #69, opened 2020-10-31
refactor: use jest test environment


danielroe:
I'd be happy to refactor the way this module is invoked to use the [Jest custom environment feature](https://jestjs.io/docs/en/next/configuration#testenvironment-string), which I think would be a more intuitive way of setting up tests.

For example, at the moment the `setupTest` approach breaks `it.only` for example, which might be non-intuitive to a person who doesn't understand what `setupTest` is doing under the hood. (Alternatively, the solution is being clearer in the documentation.)

However, because this would be a significant change, I wanted to get a :+1: from @pi0 and @ricardogobbosouza first. Thoughts?

ricardogobbosouza:
Hi @danielroe 
Sorry, but i didn't understand what you intend to do 😅

danielroe:
@ricardogobbosouza The normal jest environments are `node` and `jsdom` which provide appropriate globals for those environments.

I am proposing the a `nuxt` environment (possibly more than one).

ricardogobbosouza:
What we have at the moment is a preset, it would be incredible to have a environment for nuxt. My thoughts:
- we have to be careful that the environment is not too slow
- we also need an environment for the modules?

vinayakkulkarni:
Any chance we could use [`ava`](http://ava.li/) instead of Facebook's Jest ?

mrazauskas:
First of all, thank you for this library. After trying it out, I though it would work brilliantly as Jest environment. Happy to find this discussion.

## Working Prototype

To try out quickly how `@nuxt/test-utils` environment could work I created this [working prototype](https://github.com/mrazauskas/test-utils-environment-proposal).

That is just a quickly written code. I was trying to make only minimal changes by wrapping Jest environment module around current API. Of course, this is clumsy, but both implementations work without conflicting.

Before diving into code or details I wanted to see the whole picture. What problems the code should solve? So I was concentrating on:

- [usage examples](https://github.com/mrazauskas/test-utils-environment-proposal/tree/main/examples)
- [documentation draft](https://github.com/mrazauskas/test-utils-environment-proposal#quick-start-javascript)
- benchmark (see bellow)

These parts are the most important in the repo.

## Proposal

The test environment could improve several things:

- simplify setup. Test environment declaration lets us know the exact structure of a project. `nuxt-app` points `options.fixture` to `process.cwd()`. `nuxt-module` looks for fixture(s) inside `test/fixture` or custom directories (current default);

- `.only` works as expected;

[...]

PR #133, opened 2021-06-30
feat: jest enviroment

ricardogobbosouza:


ricardogobbosouza:
Resolve #69

ricardogobbosouza:
@pi0 have a problem on `listhen` on windows?
https://github.com/nuxt/test-utils/runs/2955816292?check_suite_focus=true#step:8:20

mrazauskas:
@ricardogobbosouza thanks for mentioning me.

Indeed I was experimenting with Jest environment for some time. Did not send a PR, because the solution did not proof to be useful. On a positive note, I found few solutions how to simplify the code base and to make it more useful. #134 is one of these simplifications / improvements. More on the way.

Back to the the environment. In general it adds two things: pragmas and global variable.

At first I was very excited about pragmas, but how to explain them to the end user? I was trying to draft some docs. On daily basis we write simply JavaScript, not pragmas. They do not look intuitive. Also there is no linting, no auto-completion or any other helpers which we are used to have. Some validation would be nice, but that will add more and more complexity. Is this really useful?

It is nice to look at pragmas, but writing them in each test file felt like programming on a notepad in some exotic language.

Global variable indirectly makes it possible to use `test.only`. Really really good. In the other hand, environment setup makes it not possible to spy on `nuxt.moduleContainer`. `jest.spyOn` is only available after the environment is installed. Some gain, some loss.

So I kept on experimenting with setup. Found a way to keep `test.only` working (without adding global variable), to be able to spy on `nuxt.moduleContainer` and much more. For instance – nesting of `setupTest` with a possibility to extend / override previous config and to roll back after. Very simple solution.

Just trying to explain why I did not send a PR after all the enthusiasm which I wrote about Jest environment.

ricardogobbosouza:
@mrazauskas It makes sense.
I think it would also confuse the user even more.
Let's see what @pi0  thinks before closing.

pi0:
Thanks for sharing your experiences @mrazauskas @ricardogobbosouza  and trying it all. 

Can we maybe still have an optional jest environment but using `this.ctx = createContext` from main utils? (instead of class attributes/methods to duplicate logic) this way we can support both. But no pressure we can do it anytime later.

Issue #88, opened 2021-02-26
How to use this package?


rwwagner90:
The docs don't go over much of the setup required to use this package. It seems to look for a nuxt.config.js file in fixture by default, but I do not understand why. Wouldn't we want to test our real apps and configs? I tried setting `rootDir` and that didn't seem to change anything.

rwwagner90:
We ended up needing to do:

```js
  setupTest({
    browser: true,
    configFile: resolve(__dirname, '../pages/fixtures/nuxt.config.js'),
    rootDir: resolve(__dirname, '../../../'),
    testDir: resolve(__dirname, '../../'),
    config: {
      rootDir: resolve(__dirname, '../../../')
    }
  });
```

I think this package should be set to use the root nuxt app as the default, so this is not required.

milenkovic:
I would completely agree with @rwwagner90. Documentation for setup is almost non-existing, and It's embarrassing to have testing utilities where even the simplest case (from documentation) is not working.

ricardogobbosouza:
Related #95

PR #134, opened 2021-07-01
refactor!: remove `testDir` and `fixture` options in favour of `rootDir`

mrazauskas:
This is an attempt to simplify public API of the library.

Currently user can point to Nuxt apps `rootDir` in several ways:

```js
setupTest({
    rootDir: process.cwd()
  })
```

Or (#127 was an attempt to change this):

```js
setupTest({
    fixture: process.cwd()
  })
```

Or:

```js
setupTest({
    testDir: process.cwd(),
    fixture: '.'
  })
```

This is puzzling and hard to document precisely. Not a surprise that one user asked: How to use this package? And came to even more confusing solution #88.

**Problem:** Three options `rootDir`, `testDir`, `fixture` do just one thing – point to `rootDir`.

It still might seem useful to have `testDir` and `fixture` for module testing. I was looking at real world usage. This is really nice idea by @danielroe to use [`examples` directory as a fixture](https://github.com/nuxt-community/html-validator-module/blob/8d5e1302f8b6b291710167edf7a73b2d322dd392/test/module.test.js#L20
) for module testing. Unfortunately here `testDir`, `fixture` do not add clarity.

### Proposal

I would suggest to remove `testDir`, `fixture` and leaving only `rootDir` option. The default value would be `'.'`.

Simple and intuitive. `rootDir` looks and works exactly as [in Nuxt](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-rootdir). Setting up test becomes the same as setting up an app (Nuxt docs: '[it is] normally not needed to configure this option'):

```js
setupTest({
    browser: true
  })
```

Setup of module test is also simple, flexible and well readable:

```js
setupTest({
    rootDir: 'test/fixtures/some-fixture'
  })

// or

setupTest({
    rootDir: 'examples/some-example'
  })
```

By the way, `resolveRootDir` (#107) was solving the same problem. Essentially it does one important thing – set the default value `options.rootDir = '.'`. I just simplified the solution.


ricardogobbosouza:
Nice @mrazauskas

ricardogobbosouza (approved):

pi0:
@mrazauskas Thanks for the PR. Seems much better with simplified options. Would you please rebase it too? (I will double check DX locally for this once before merge if we can still have one more conventional opetion to set root but otherwise using `rootDir` is fine)

mrazauskas:
@pi0 Thanks. Sure, I will rebase it later today.

mrazauskas:
@ricardogobbosouza Could I ask to revert? I just wanted to improve a couple of details, it took longer than I was expecting, but it would be good to include them with this PR. Thanks!

PR #136, opened 2021-07-06
fix: close server explicitly

mrazauskas:
After migration to `listhen` (#129), server is not closed after tests run. Therefore, Jest throws “A worker process has failed to exit gracefully…” error. To recreate, try building the library and importing `setupTest` to a test from `dist`.

`listhen` has close on exit signal feature, but for some reason it does not work in Jest environment. The test below will throw the same error, because listener is not closing:

```js
import { listen } from 'listhen'

test('listhen', async () => {
  await listen((req, res) => { res.end(req.url) })
  // ...
})
```

The idea of auto closing server inspired me and I thought perhaps Nuxt could be also closed on exit signal. Did not manage to make this work and also found out that server is not closing. (I was trying out 'beforeExit' event too. No luck.)

Seems like everything should be closed explicitly.

### DIFF
--- a/src/setup.ts
+++ b/src/setup.ts
@@ -21,6 +21,10 @@ export function setupTest (options: Partial<NuxtTestOptions>) {
       await ctx.nuxt.close()
     }
 
+    if (ctx.listener) {
+      await ctx.listener.close()
+    }
+
     if (ctx.browser) {
       await ctx.browser.close()
     }
--- a/src/context.ts
+++ b/src/context.ts
@@ -2,6 +2,7 @@ import type { RequestListener } from 'http'
 import { join } from 'path'
 import defu from 'defu'
 import { NuxtConfig, NuxtOptions } from '@nuxt/types'
+import type { Listener } from 'listhen'
 import type { Browser, LaunchOptions } from 'playwright'
 
 export interface NuxtTestOptions {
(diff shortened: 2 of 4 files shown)


src/setup.ts
```diff
@@ -17,13 +17,9 @@ export function setupTest (options: Partial<NuxtTestOptions>) {
...
-    if (ctx.nuxt) {
-      await ctx.nuxt.close()
-    }
-
-    if (ctx.browser) {
-      await ctx.browser.close()
-    }
+    await ctx.browser?.close()
+    await ctx.listener?.close()
+    await ctx.nuxt?.close()
```
ricardogobbosouza, line 22: I have doubts in the closing sequence.
I believe the browser should be the last, no?

ricardogobbosouza (commented):
Nice @mrazauskas, just a question at closing.

src/setup.ts
mrazauskas: Good question. My thinking was: browser might depend on server and server might depend on something what was created with nuxt instance. In theory last-in-first-out sequence makes sense. I was testing the opposite too, but could not proof any difference. So it could be left how you prefer.

mrazauskas:
@ricardogobbosouza Thanks for the question. It sparkled one idea, which solved another problem I was struggling with.

Closing sequence is reverted back. There is no need to change it.

A unit test is added. Simply for documentation.

pi0 (approved):
Thanks!

PR #137, opened 2021-07-07
feat!: add `randomBuildDir` option

mrazauskas:
It would be useful to have a possibility to disable random `buildDir` explicitly.

Current:
* `buildDir` is patched with random string by default, but it will be implicitly not patched if custom `buildDir` is set inside `nuxt.config` file or `config` overrides has `buildDir` value
* there is no possibility to disable random default
* there is no possibility to add random part for a custom `buildDir`

Proposed:
* `buildDir` is always randomised (predictable)
* random part will be disabled if `randomBuildDir: false` is set (explicit)

**Use case.** If testing an app, one would be able to build it using a `package.json` script and run tests with `build: false` option set inside `setupTest`. There is no need to build the app with each `setupTest`, this makes the tests suite run significantly faster.

Also proposed implementation is easier to document and to use, because of explicit configuration and predictable behaviour.

**Breaking.** If this PR is merged, previous behaviour can be achieved like this:
```js
setupTest({
  build: true,
  randomBuildDir: false
})
```

**Testing, debugging.** `randomId` was factored out for better testability. `Date.now()` segment was added to make random directories sortable. This helps to find the recent build easier (useful for debugging).

**Types.** `options` object is always partial.

### DIFF
--- a/docs/content/en/api-reference/setup.md
+++ b/docs/content/en/api-reference/setup.md
@@ -71,21 +71,12 @@ The path to the Nuxt application that will be used in the tests.
 * Type: `string`
 * Default: `<testDir>/<fixture>`
 
-### Build directory
+##### randomBuildDir
 
-`setupTest` will create a random build directory to avoid race conditions and conflicts between tests which run in parallel. The default template is: `<rootDir>/.nuxt/<randomID>`.
+To avoid conflicts between concurrent tests, a new random [build directory](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-builddir) will be created each time `setupTest` is called.
 
-If `buildDir` is set in `nuxt.config`, the value will be used instead of random one.
-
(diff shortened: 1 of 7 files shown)


ricardogobbosouza:
Nice @mrazauskas!
I think this PR doesn't break anything.

ricardogobbosouza (approved):

mrazauskas:
@ricardogobbosouza In very rare cases it is breaking. For instance if a test was in some way relying on implicitly not randomised custom `buildDir`. Now `randomBuildDir: false` must be set explicitly.

pi0 (approved):

pi0:
@mrazauskas Would you please resolve conflicts?

mrazauskas:
@pi0 Sure. Just rebased everything. Thanks for accepting the ideas.

PR #138, opened 2021-07-08
feat!: add `randomPort` option

mrazauskas:
It would be also useful to have `randomPort` option. Motivation is similar to `randomBuildDir` #137

Current:
* test server is trying to listen on `config.server.port`, random port is a fall back
* it is not possible to choose explicitly which value should be used in a test

Proposed:
* server always listens on random port (predictable)
* if `randomPort: false` is set, custom port is used (explicit)

**Use case.** For example, a fixture could have `config.server.port` set. One test ignores that. Another test has `randomPort: false` set and tries to listen on that particular port. There is no need to have separate fixtures or any additional logic. Inspired by real use case nuxt-community/auth-module#868 (see test file).

**Breaking.** If this PR is merged, previous implicit behaviour gets broken. Easy to fix it:
```js
setupTest({
  server: true,
  config: { server: { port: 5050 } },
  randomPort: false
})
```

**Types.** `options` object is always partial. Factored out `Nuxt` interface is useful in test.

### DIFF
--- a/docs/content/en/api-reference/setup.md
+++ b/docs/content/en/api-reference/setup.md
@@ -87,6 +87,14 @@ setupTest({
 })
 ```
 
+##### randomPort
+
+The test server will listen on a new random port each time `setupTest` is called. If disabled, the server will try to use [`server.port`](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-server) value.
+
+* Type: `boolean`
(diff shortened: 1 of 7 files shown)


ricardogobbosouza (approved):

mrazauskas:
Rebased after #136 was merged.

pi0 (approved):

PR #141, opened 2021-07-09
build: externalise `ohmyfetch/node`

### LINKED ISSUE #140: Build size
Just noticed that for some reason `ohmyfetch` is dramatically increasing the build size:
* if `ohmyfetch` is commented out in the code: 7.5 kB
* otherwise: **52.1 kB**

Is this intentional?

danielroe:
resolves #140

### DIFF
--- a/siroc.config.ts
+++ b/siroc.config.ts
@@ -1,0 +1,7 @@
+import { defineSirocConfig } from 'siroc'
+
+export default defineSirocConfig({
+  rollup: {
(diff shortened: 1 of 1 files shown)


pi0:
I'm pretty sure saw same error with imports with path. Since `ohmyfetch` is already in `dependencies`, I think this is mostly a siroc bug :)

danielroe:
@pi0 Good point, I'll fix there.

PR #142, opened 2021-07-09
chore: dependency clean up

### LINKED ISSUE #140: Build size

danielroe:
* move babel dev deps required by `jest-preset.js` to dependencies: see #93
* upgrade all dependencies & dev dependencies: the `siroc` upgrade resolves #140 
* pin dev deps so they can be managed by @renovate-bot in future

Happy to close this or split PR up if different approach is preferred (cc: @pi0).

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,29 +19,29 @@
     "test:unit": "jest test/unit --forceExit"
   },
   "dependencies": {
-    "@babel/preset-typescript": "^7.13.0",
+    "@babel/plugin-transform-runtime": "^7.14.5",
+    "@babel/preset-env": "^7.14.5",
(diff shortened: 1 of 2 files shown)


mrazauskas:
Just thinking out loud – is it really useful to include `babel-jest` as dependency of this library? Makes me doubt, because the actual copy of `jest` installed by the user will depend on `babel-jest`. For instance, `jest` v26 bundles and depends on `babel-jest` v26.

PR #143, opened 2021-07-12
refactor: improve nuxt app validation

mrazauskas:
While rebasing #134, I though the Nuxt app validation could be simplified.

Currently: always three calls to the file system.
Proposed: just one `readdir` call and validation in memory. Array’s `.some()` should help to return early.

In some exotic case, `pages` could also be a file. It is possible to detect this using `Dirent`s, but perhaps that’s too much for this small library.

Validation logic is factored out to improve testability.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,22 @@
+import { readdir } from 'fs/promises'
+import { resolve } from 'path'
+
+const nuxtDir = ['nuxt.config.js', 'nuxt.config.ts', 'pages']
+const fixtureError = 'Make sure \'rootDir\' or \'fixture\' option is pointing to a location which at least contains \'nuxt.config\' file or \'pages\' directory.'
+
+export const ensureNuxtApp = async (dir: string) => {
+  const targetDir = await readdir((resolve(dir)))
+    .catch((_err) => {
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -1,3 +1,22 @@
+import { readdir } from 'fs/promises'
+import { resolve } from 'path'
+
+export const isNuxtAppDir = async (dir: string) => {
```
pi0, line 4: Since function is not returning anything but ensuring, I think would be better calling it `ensureNuxtAppDir`
```suggestion
export async function ensureNuxtAppDir (dir: string) {
```

src/utils.ts
```diff
@@ -1,3 +1,22 @@
...
+  const fixtureError = 'Make sure \'rootDir\' or \'fixture\' option is pointing to a location which at least contains \'nuxt.config\' file or \'pages\' directory.'
+  const nuxtDir = ['nuxt.config.js', 'nuxt.config.ts', 'pages']
```
pi0, line 6: We can move consts outside of function

src/utils.ts
mrazauskas: And perhaps to make it shorter – `ensureNuxtApp(dir: string)`. Clear enough?

pi0:
Thanks!

PR #144, opened 2021-07-12
chore(deps): remove `babel-jest`

mrazauskas:
`babel-jest` is `jest`'s dependency. `jest` takes care to install and makes sure to pick correct version. It is not possible to know in advance if user has Jest v26 or v27.

Might be there is no harm to declare it as a dependency, but there is no need or gain too. This library does not import from `babel-jest`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "@babel/plugin-transform-runtime": "^7.14.5",
     "@babel/preset-env": "^7.14.5",
     "@babel/preset-typescript": "^7.14.5",
-    "babel-jest": "^27.0.6",
     "consola": "^2.15.3",
     "defu": "^5.0.0",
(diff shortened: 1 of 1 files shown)


ricardogobbosouza:
Hi @mrazauskas 
We need this dependency in https://github.com/nuxt/test-utils/blob/main/jest-preset.js#L6

mrazauskas:
@ricardogobbosouza Yes, it is included in the preset, but the preset will be loaded by a copy of Jest installed by the user. If the user picks Jest v26, `babel-jest` v26 will be needed. Jest will be always better in managing its dependencies.

Don’t know how to explain it better.

ricardogobbosouza:
We are testing other transpilers as babel is taking too long.
#145

PR #145, opened 2021-07-12
chore: use @sucrase/plugin

ricardogobbosouza:
I did some tests using [`@sucrase/plugin`](https://github.com/alangpierce/sucrase/tree/main/integrations/jest-plugin) and it turned out to be much faster than babel

It also managed to transpile the `node_modules` folder, but it took a little longer.

Maybe it will solve #132, just adding :

```js
transformIgnorePatterns: []
```

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -1,16 +1,5 @@
 module.exports = {
   collectCoverage: true,
   testEnvironment: 'node',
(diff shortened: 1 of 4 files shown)


ricardogobbosouza:
What do you think? @pi0

PR #147, opened 2021-07-13
test: improve `ensureNuxtApp` test

mrazauskas:
@ricardogobbosouza Thanks for solution. It is better without `@ts-ignore`.

Just double checked, in this case `readdir` will return an array. `string[]` here looks better for me.

### DIFF
--- a/test/unit/utils.test.ts
+++ b/test/unit/utils.test.ts
@@ -5,20 +5,19 @@ const mockReaddir = readdir as jest.MockedFunction<typeof readdir>
 
 jest.mock('fs/promises')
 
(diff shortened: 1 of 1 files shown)


ricardogobbosouza:
@mrazauskas thanks
I had unintentionally done it in another PR

ricardogobbosouza (approved):

mrazauskas:
Just did few more tiny improvements. Here is the whole list:

* removed unnecessary `@ts-ignore`
* corrected name of the `ensureNuxtApp` test
*  simplified mocking

PR #148, opened 2021-07-13
refactor: move types to separate file

mrazauskas:
Would you agree to move types to separate file?

They are used in tests now. It looks logic to import types from `'src/types'`.

And I work one improvement for the context. This will be slightly larger refactor of test setup and context logic. It looks better structured with types moved to a separate file. This PR also makes the next PR slimmer and both are easier to review. Hope this makes sense.

Few changes (mostly clean up):
* for consistency methods are always typed `ready(): any` (instead of `ready: () => any`). Can be different style if you prefer.
* return value of `ready`, `close` and `build` is corrected to `Promise<>`.


ricardogobbosouza:
@mrazauskas nice!

ricardogobbosouza (approved):

PR #150, opened 2021-07-14
refactor: launch server on demand

mrazauskas:
To enable helpers like `get` or `createPage`, documentation is reminding to set `{ server: true, browser: true }`.

`get` throws an error if server is not listening. The logic inside `browser.ts` is different, [a browser is created](https://github.com/nuxt/test-utils/blob/9b239ac1550118262440c20805ceca6fc69c14fc/src/browser.ts#L29) if it is not found in the context.

This inconsistence was bugging me for some time. What if both (server or browser) would be launched automagically then some helpers requires them? The helpers would be much more helpful, isn’t it? It is obvious what user is expecting:

```js
setupTest()
test('some test', async () => {
  const html = await get('/')

////

setupTest()
test('some test', async () => {
  const page = await createPage('/')
```

One tiny problem – `build: true` should be set to make this work. Exchanging `server: true` for `build: true` does not look like any improvement. I think defaults should be adjusted. See #151.

PS Tests will work only after #151 is merged. I made it like this on purpose.

### DIFF
--- a/docs/content/en/api-reference/server-testing.md
+++ b/docs/content/en/api-reference/server-testing.md
@@ -8,8 +8,6 @@ category: Writing tests
 
 Nuxt test utils exposes a number of useful methods you can use when testing a Nuxt application.
 
-These helper methods require that you pass `{ server: true }` as an option to `setupTest` ([more info](/api-reference/setup#features-to-enable)).
-
 ## get
 
 You can get a response to a server-rendered page with `get`.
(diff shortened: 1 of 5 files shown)


ricardogobbosouza:
@mrazauskas nice
It may no longer make sense to have `server` and `browser` as options.
When necessary it will be automatically started, this helps the user.

https://github.com/nuxt/test-utils/blob/main/src/setup.ts#L48
https://github.com/nuxt/test-utils/blob/main/src/setup.ts#L60

mrazauskas:
> It may no longer make sense to have `server` and `browser` as options.

@ricardogobbosouza Might be you are right. That’s very good insight. If helpers do the job, users will not use these options at all. Would be good to tryout edge cases, for instance what happens if we have `build: false` set? I will check.

ricardogobbosouza:
We can initialize the build when needed and store it in context

PR #151, opened 2021-07-14
refactor!: change defaults `build: true`

mrazauskas:
`build: true` might be better default:

One. It would make helpers more helpful (see #150).

Two. In general `setupTest` utility is helping to set up end-to-end tests. The artefacts created by building or generating apps/fixtures are the actual fixtures which are being put under tests. Without `build` step there is nothing to test. I see only two exceptions: a) the user might build the app before launching test, in this case `build: false` is necessary also with current defaults; b) the `moduleContainer` assertions are used (these work without `build` step).

The `moduleContainer` assertions work with `build: true`. It is only the matter of speed. `build: false` would make them fast again. Perhaps this could be the solution for now. I have better idea, but that’s another PR.

~~`generator` logic got improved a bit. I was checking if there are no side effects. If builder is created by default, better to reuse it.~~

### DIFF
--- a/docs/content/en/api-reference/browser-testing.md
+++ b/docs/content/en/api-reference/browser-testing.md
@@ -8,8 +8,6 @@ category: Writing tests
 
 Under the hood, Nuxt test utils uses [`playwright`](https://playwright.dev/) to carry out browser testing.
 
-These helper methods require that you pass `{ browser: true }` as an option to `setupTest` ([more info](/api-reference/setup#features-to-enable)).
-
 ## createPage
 
 You can initiate a browser session for a given page with `createPage`.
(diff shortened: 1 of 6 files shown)


ricardogobbosouza (approved):

mrazauskas:
Found one problem. `build` step was also running with `generate: true`. It might create some problems. Disabling `build` if `generate: true` should work similar to `target: static`. Right?

mrazauskas:
Found one problem. The `build: true` default makes `build` step also run with `generate: true`. This overlaps with `generate` step and might create problems.

I mean, `build: true` is the same as `target: 'server'` (runs `build` step). And disabling `build` step if `generate: true` should work similar to `target: 'static'`.

PR #152, opened 2021-07-14
refactor: reuse builder and build logic

mrazauskas:
@ricardogobbosouza Inspired by your [comment](https://github.com/nuxt/test-utils/pull/150#issuecomment-879898834)

Generator could reuse builder if found in the context. And could add it to context if not found.

Or am I overthinking it and current use of `new Builder`, `new Generator` is more optimal?


mrazauskas:
Closing. Just fixed the overlap of functionality in #151.

PR #597, opened 2023-12-02
fix: teardown test build directories

harlan-zw:
Review this first: https://github.com/nuxt/test-utils/pull/596

## Issue

Test build dirs will just keep stacking up. There is no logic in place to remove them.

This isn't ideal, especially when your IDE tries to index these files (even though they're ignored - phpstorm :angry:)

image

## Solution

This PR includes a few important fixes:
- group test builds in a `test` directory so they can easily be cleaned up manually, this may be needed if a test fails/is force quit
- mark the build directory that's created as a side effect and clean it up once the tests are finished

### DIFF
--- a/src/core/nuxt.ts
+++ b/src/core/nuxt.ts
@@ -43,7 +43,7 @@ export async function loadFixture () {
 
   if (!ctx.options.dev) {
     const randomId = Math.random().toString(36).slice(2, 8)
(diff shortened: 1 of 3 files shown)


danielroe (commented):
This is much needed, thank you!

What about calling it teardown or cleanup? `sideEffects` doesn't entirely communicate to me that these will run after the test finishes

danielroe (approved):
❤️

Issue #650, opened 2023-12-15
in-source testing support


danielroe:
Looks like there might be a regression with [in-source testing](https://vitest.dev/guide/in-source.html#typescript):

https://github.com/logotip4ik/keycap/pull/16

PR #651, opened 2023-12-15
feat(vitest-environment): add support for in-source testing

danielroe:
resolves https://github.com/nuxt/test-utils/issues/650

We get this error: https://github.com/vitest-dev/vitest/issues/3393

cc: @sheremet-va in case you have any ideas.


logotip4ik:
@danielroe I might be mistaken, but you probably mentioned wrong person. Here is handle for Vitest maintainer (Vladimir): `sheremet-va`

sheremet-va:
In-source tests assume that code is running when the file is imported. So, in the example that I see, I am not sure that the code is actually running since the component needs to be mounted, right?

danielroe:
Thank you !!

We might consider hoisting in-source testing out of the setup function later on.

danielroe:
Linking https://github.com/microsoft/TypeScript/issues/45096 - which was the cause of CI failure.

Worked around for the moment.

PR #670, opened 2023-12-21
[WIP] feat: add setRuntimeConfig module utility using IPC signal

BobbieGoede:
This is an exploration on how to update/set runtime config within a running server process, see https://github.com/nuxt/test-utils/pull/640 for more information.


alexandre-hallaine:
maybe we have to close it

Issue #585, opened 2023-07-21
mountSuspended with i18n


ronaldkonjer:
I'm setting up the nuxt-vitest module in my project. Though I'm running into some issues getting a simple test, I use mountSuspended(Logo) to test the mounting of a Logo component. This component uses i18n. Whereas when I'm using mount from Vitest the test passes.

The error I'm getting is as follows:
```
Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
SyntaxError: Need to install with `app.use` function
 ❯ Module.createCompileError ../node_modules/.pnpm/@intlify+message-compiler@9.3.0-beta.17/node_modules/@intlify/message-compiler/dist/message-compiler.mjs:54:19
 ❯ createI18nError ../node_modules/.pnpm/vue-i18n@9.3.0-beta.17_vue@3.3.4/node_modules/vue-i18n/dist/vue-i18n.runtime.mjs:97:34
     96|     [I18nErrorCodes.INVALID_ARGUMENT]: 'Invalid argument',
     97|     [I18nErrorCodes.MUST_BE_CALL_SETUP_TOP]: 'Must be called at the top of a `setup` function',
     98|     [I18nErrorCodes.NOT_INSLALLED]: 'Need to install with `app.use` function',
       |                                  ^
     99|     [I18nErrorCodes.UNEXPECTED_ERROR]: 'Unexpected error',
    100|     [I18nErrorCodes.NOT_AVAILABLE_IN_LEGACY_MODE]: 'Not available in legacy mode',
[...]

PR #673, opened 2023-12-22
Update i18n example test to actually exercise vuei18n

sarayourfriend:
This shows a minimal reproduction of the issue discussed in #585


examples/i18n/app.vue
```diff
@@ -1,5 +1,14 @@
 <template>
   <div>
-    Hi from @nuxtjs/i18n: {{ (('hello')) }}
+    Hi from @nuxtjs/i18n: {{ $t('hello') }}
+  </div>
+  <div>
```
obulat, line 5:
```suggestion
  <i18n-t scope="global" keypath="message"><template #module>@nuxtjs/i18n</template></i18n-t>
  <div>
```
The error in i18n _can_ be fixed by importing `i18n` from `useNuxtApp().$i18n`. However, I couldn't find a way to fix the `i18n-t` component. I tried replacing it with the `Translation` component imported from `vue-i18n`, but the error persists.

This suggested change will also need an addition to the `i18n` messages:
`"message": "Hello from {module}"`

danielroe:
As far as I can tell the issue here is because the `@nuxtjs/i18n` module is calling `vueApp.use` on the CJS version of `vue-i18n`, but when we use `useI18n` in the test, it's the ESM version of the library, and so it thinks it's not registered.

Without looking further (though of course I will) I'm not sure whether this is something that `@nuxt/test-utils` should solve or whether it's an upstream issue ....

cc: @BobbieGoede so you're aware

sarayourfriend:
Thanks for figuring that out! Indeed, that doesn't seem like something this library should fix. @obulat I wonder if we can force Openverse (or at least the test suite) to use the CJS version?

obulat:
I opened an issue with the reproductions (using composable and `i18n-t` component) in the `@nuxt-modules/i18n` repository.
@sarayourfriend, I tried to find a way to hot-fix this issue, but I couldn't. 
@danielroe, thank you for investigating this issue. Do you have any suggestions for hot-fixing this until the solution is found?

danielroe:
I still plan to investigate this further.

indiehjaerta:
Any updates on the investigation of this? As it stands now this problem stops us from upgrading to the  @nuxt/test-utils from nuxt-vitest that we have used before.

Issue #686, opened 2024-01-04
RouterLink not using correct href


StevenPewsey:
### Environment

```
------------------------------
- Operating System: Darwin
- Node Version:     v18.16.1
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  yarn@3.6.0
- Builder:          -
- User Config:      app, css, build, devServer, modules, runtimeConfig, nitro, vite, i18n, hooks
- Runtime Modules:  @nuxtjs/tailwindcss@6.10.1, @pinia/nuxt@0.4.11, @pinia-plugin-persistedstate/nuxt@1.2.0, @thrivadev/thrive/nuxt@2.1.1, @vueuse/nuxt@10.7.0, @nuxtjs/i18n@8.0.0-rc.9, @vee-validate/nuxt@4.12.2
- Build Modules:    -
------------------------------
```

### Reproduction

https://stackblitz.com/edit/github-tqbkqz

### Describe the bug

The mocked `RouterLink` [component](https://github.com/nuxt/test-utils/blob/main/src/runtime-utils/components/RouterLink.ts) is passing `props.to` directly as the href, however this results in the href being an object when `to` is an object (e.g. if passing through `path` and `state`).

This causes tests to fail even though the links work fine in the browser

```
- http://localhost:3000/thelink
+ http://localhost:3000/[object%20Object]
```

manniL:
Indeed! PR welcome to address that issue so the whole path is passed if `props.to` is no string but a route object 👍🏻

PR #687, opened 2024-01-04
fix(runtime-utils): RouterLink - use the correct href when props.to is object

StevenPewsey:
## What
* Update RouterLink to make the `to` prop required (it's required in the actual RouterLink component https://github.com/vuejs/vue-router/blob/0395bd83feb79b3cba399e68f991885c986f70ec/src/components/link.js#L23)
* Get the href from the resolved route, rather than `props.to`

## Why
* Fixes https://github.com/nuxt/test-utils/issues/686


danielroe (approved):
❤️

PR #726, opened 2024-01-19
fix(runtime): reset modules after nuxt setup runs

danielroe:


examples/app-vitest-full/tests/setup/mocks.ts
```diff
@@ -1,6 +1,8 @@
 import { vi } from 'vitest'
 import { mockNuxtImport } from '@nuxt/test-utils/runtime'
 
+vi.resetModules()
```
huang-julien, line 4: is it mandatory ?

danielroe: yes, without it vitest errors and tells us to reset these mocks...

huang-julien: So i guess we should warn users that this is needed within setup files ? Or probably inject it ourselves with a transform plugin ?

danielroe: I'm thinking we shouldn't insert it ourselves as they might want to run some setup steps that rely on not resetting modules. In future we could add it as an enhancement (or bugfix!) if there's not a reason to allow manually calling it.

Issue #732, opened 2024-01-23
Server utils composable is not defined in test


PoloLacoste:
### Environment

- Operating System: Windows_NT
- Node Version:     v21.6.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.14.3
- Builder:          -
- User Config:      modules, devtools
- Runtime Modules:  @nuxt/test-utils/module@3.10.0
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-7fr8te?file=test%2Fapp.test.ts

### Describe the bug

In`server/domain` I created a `TestService` class which is importing a composable from `server/utils`.

`server/utils/test.ts` :
```ts
export const useTest = () => {
  return 1
}
```

`server/domain/test-service.ts` :
```ts
export class TestService {
  private test = useTest()
}
```

When I initialize the `TestService` class inside an api endpoint it works, but when I try to initialize the same class inside a test this error appears `ReferenceError: useTest is not defined`.

`test/app.test.ts` :
```ts
// @vitest-environment nuxt
import { expect, test } from 'vitest';
import { TestService } from '~/server/domain/test-service';

test('my test', () => {
  const service = new TestService();
  expect(service).toBeDefined();
});
```

danielroe:
Nitro server environment is not yet supported - the environment provided is a _client-side_ runtime vitest environment.

You can track this feature in https://github.com/nuxt/test-utils/issues/531.

Issue #677, opened 2023-12-26
Nuxt module import in E2E environment


peterbud:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.10.5
- Builder:          -
- User Config:      modules, devtools, runtimeConfig
- Runtime Modules:  @nuxt/test-utils/module@3.9.0, ()
- Build Modules:    -
------------------------------

### Reproduction

With the latest nuxt test-utils I was trying to mock a server side import. Before even able to mock the server side import, I have realized that the plain vanilla minimal nuxt module seems to be ignored/not executed when running e2e test. 

See repro here:
https://stackblitz.com/edit/nuxt-vitest-x6ycht

### Describe the bug

Basic setup is to have a server side import coming from a module. an API is using that import to return a value. Running `pnpm dev` works, and the API returns the value as expected.

Running an E2E test in the nuxt environment however fails, showing error messages which are hinting that the module import has not been executed:

`(node-resolve plugin) Could not resolve import "#auth" in /home/projects/nuxt-vitest-x6ycht/server/api/test.get.ts using imports defined in /home/projects/nuxt-vitest-x6ycht/package.json.`

Maybe my test config needs to be improved? Or is it a bug?

danielroe:
Currently the vitest environment only allows for testing _client-side_ code. (So any aliases added to your nitro config are not respected.)

We can track supporting server side unit testing in https://github.com/nuxt/test-utils/issues/531.

peterbud:
Thanks Daniel. I had the impression that E2E is also including the Nuxt server component. Thanks for the clarification.

danielroe:
You can definitely test the server side with e2e, but that starts a server with `setup` and performs fetch requests to it. You don't directly import the server code into your test.

Issue #531, opened 2023-06-02
nuxt server environment support


danielroe:
We currently support running composables/components in a browser-type environment (with happy-dom, and soon jsdom). But there are use cases where it might be useful to support running tests in a hybrid server environment where things like h3 utilities and nuxt ssr utilities work.

This could definitely be classed as non-essential and experimental, but we can track it here.

alexcroox:
Are there any plans to make this happen still? I was hoping for my full stack Nuxt framework that I'd be able to unit test the full stack but the confusion/limited support for unit testing the server side has resulted in weeks of lost time and confusion around the matter. Even a section in the testing docs warning people away from unit testing the server/api routes would help others here.

zoobzio:
For unit tests on Nuxt API endpoints this problem might also be solved w/ H3 test utils that are agnostic of the Nuxt runtime & instead allow for programmatic execution of event handlers 

I have been playing w/ an implementation that allows for me to:

- mock `h3` utils like `defineEventHandler` & automatically stub the global context 
- pass mocked `H3Event` data to a given handler function as a test cases 

Extending [this workaround](https://github.com/nuxt/test-utils/issues/531#issuecomment-1836960466), we can define a function that registers `h3` mocks in the global context:

```ts
// ~/test/setup.ts
import type { H3Event } from "h3";
import { vi } from "vitest";

type Handler = (event: H3Event) => Promise<unknown>;

export function useH3TestUtils() {
  const h3 = vi.hoisted(() => ({
    defineEventHandler: vi.fn((handler: Handler) => (event: H3Event) => handler(event)),
    // TODO mock these utils: https://www.jsdocs.io/package/h3#package-index-functions 
  }); 
 
  // stub `h3` & functions individually to support auto-imports
  vi.stubGlobal("h3", h3);
  vi.stubGlobal("defineEventHandler", h3.defineEventHandler);
  // TODO stub remaing utils

  return h3;
}
```

This makes unit testing a given handler a breeze:

```ts
// ~/test/server/api/handler.test.ts 
import { useH3TestUtils } from "@@/test/setup";
import { describe, it, expect } from "vitest";

const { 
  defineEventHandler, 
  // TODO add mocked utils your handler uses... 
} = useH3TestUtils();

describe("my handler", async () => {
  const handler = await import("../../../server/api/handler.get.ts");

  it("is registered as an event handler", () => expect(defineEventHandler).toHaveBeenCalled());
  
  it("does something", async () => {
[...]

PR #757, opened 2024-02-14
feat: add unimport plugin for Nitro support

productdevbook:
cc: https://github.com/nuxt/test-utils/issues/531, https://github.com/nuxt/test-utils/issues/732, https://github.com/nuxt/test-utils/issues/677

This supports automatic imports and aliases on the server side.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,7 @@ import type { DotenvOptions } from 'c12'
 import type { InlineConfig } from 'vite'
 import { defu } from 'defu'
 import { createResolver } from '@nuxt/kit'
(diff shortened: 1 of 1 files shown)


danielroe:
Thank you for this.

We probably need this to be an entirely separate vitest environment, as there will be a lot of what the current vitest environment does that shouldn't be done for nitro/server routes.

productdevbook:
I don't know how to do it the other way around, so don't expect a pr from me. :D But this pr is workable for testing composables like utils etc.

I hope we will soon have the ability to test this nitro.

PR #777, opened 2024-03-14
chore: Playwright e2e testing enhancements

mxschmitt:


examples/app-playwright/package.json
```diff
@@ -16,7 +16,7 @@
     "vue-router": "^4.2.5"
   },
   "devDependencies": {
-    "playwright": "^1.42.1",
-    "@nuxt/test-utils": "latest"
+    "@nuxt/test-utils": "latest",
+    "@playwright/test": "1.43.0-alpha-2024-03-14"
```
mxschmitt, line 20: Note: thats the alpha version with the fix:  https://github.com/microsoft/playwright/pull/29865

examples/app-playwright/playwright.config.ts
```diff
@@ -1,3 +1,29 @@
+import { fileURLToPath } from 'node:url'
 import { defineConfig } from '@playwright/test'
+import type { ConfigOptions } from './customFixtures'
 
-export default defineConfig({})
+const nuxt = {
+  rootDir: fileURLToPath(new URL('.', import.meta.url)),
+}
+
+export default defineConfig<ConfigOptions>({
```
mxschmitt, line 9: Thought: Ideally we mirror here a similar default config like we have here: https://github.com/microsoft/create-playwright/blob/main/assets/playwright.config.ts

We can strip a lot of unrelevant stuff from it tho.

examples/app-playwright/test/basic.test.ts
```diff
@@ -1,14 +1,6 @@
...
-  rootDir: fileURLToPath(new URL('../', import.meta.url)),
-  // TODO: auto-detect playwright
-  runner: 'playwright'
-})
+import { expect, test } from '../customFixtures'
 
 test('test', async ({ page }) => {
-  await page.goto(url('/'))
-  test.expect(await page.getByRole('heading').innerText()).toBe('Welcome to Playwright!')
+  await page.goto('/')
```
mxschmitt, line 4: Instead of relying on the `url()` function from the test-utils package, we use Playwright's baseURL.

examples/app-playwright/test/basic.test.ts
```diff
@@ -1,14 +1,6 @@
...
+  await expect(page.getByRole('heading')).toHaveText('Welcome to Playwright!')
```
mxschmitt, line 5: These assertions will automatically retry over and over again until the condition is met or the timeout has been reached (they are non-flaky).

danielroe (approved):

PR #781, opened 2024-03-15
feat: expose goto() fixture with hydration features

mxschmitt:


examples/app-playwright/tests/basic.test.ts
```diff
@@ -10,7 +10,7 @@ test.use({
 })
 */
 
-test('test', async ({ page }) => {
-  await page.goto('/')
+test('test', async ({ page, goto }) => {
+  await goto('/', { waitUntil: 'hydration' })
```
danielroe, line 14: would this behaviour also work with `page.goto`?

danielroe (approved):

examples/app-playwright/tests/basic.test.ts
mxschmitt: Unfortunately not, we don't recommend overriding page.goto so I think exposing it as a fixture is a nice middle-line between both sides (overriding class vs. making this extra functionality available).

PR #785, opened 2024-03-16
ci: remove caching of Playwright browsers

mxschmitt:
Caching the Playwright browsers is a concept which often leads to confusion, since yes, browsers can be cached, but they also rely on operating system dependencies, which can't be cached.

This might work for Chromium, since GitHub Action environments have all the OS dependencies already pre-installed for running Chromium, but for WebKit / Firefox this won't work (launch will yell with dependencies missing error).

So having this extra complexity for saving 1-5 seconds on a process (downloading the browsers) which only [takes 5 seconds](https://github.com/nuxt/test-utils/actions/runs/8303902994/job/22728801115#step:9:11) seems not worth it (restoring the cache takes a similar duration)

See also here: https://playwright.dev/docs/ci#caching-browsers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,6 @@ env:
   # 7 GiB by default on GitHub, setting to 6 GiB
   # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
   NODE_OPTIONS: --max-old-space-size=6144
-  # install playwright binary manually (because pnpm only runs install script once)
-  PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: "1"
 
 
 # Remove default permissions of GITHUB_TOKEN for security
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -13,8 +13,6 @@ env:
   # 7 GiB by default on GitHub, setting to 6 GiB
   # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
   NODE_OPTIONS: --max-old-space-size=6144
-  # install playwright binary manually (because pnpm only runs install script once)
-  PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: "1"
```
mxschmitt, line 17: drive-by: Playwright does not install browsers as part of `install script` [since v1.38](https://playwright.dev/docs/release-notes#version-138).

danielroe (approved):
Thank you!

Issue #314, opened 2022-10-19
The "setup" in the tests is very slow for every rerun


enkot:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.12`
- Nitro Version:    `0.6.0`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-framework-zehqm6

### Describe the bug

The empty `await setup({})` in the tests takes ~4 seconds for every rerun in `watch` mode in the empty project.
Please, check the reproduction, try to run `npm test`.

nWacky:
> Is it possible to have a dev. instance running & to use that as the "server" for the e2e tests?

I managed to avoid rebuilding every time by building into `.output/output` 

```bash
npm run build && mv .output .output/output
```

And then setting output directory to `.output`

```ts
// playwright.config.ts
export default defineConfig<ConfigOptions>({
  use: {
    nuxt: {
      build: false,
      buildDir: ".output",
      rootDir: fileURLToPath(new URL(".", import.meta.url)),
    },
  },
})
```

For now nuxt always sets [`nitro.output.dir` to `buildDir/output`](https://github.com/nuxt/test-utils/blob/ba1fa2d91bd7db6102af9c3a7048745189f39086/src/core/nuxt.ts#L51) and [starts a server from `buildDir/output`](https://github.com/nuxt/test-utils/blob/ba1fa2d91bd7db6102af9c3a7048745189f39086/src/core/server.ts#L54)

`npx playwright test` is using an existing build after this

linspw:
You are amazing @nWacky !
I only adjusted to doesn't need mv command:

```ts
export default defineConfig<ConfigOptions>({
  use: {
    nuxt: {
      build: false,
      buildDir: '.output',
      nuxtConfig: {
        nitro: {
          output: {
            dir: '.output',
          },
        },
      },
      rootDir: fileURLToPath(new URL('.', import.meta.url)),
    },
})
```

linspw:
Wrapping nuxt setup:
Be Happy

OBS: The test now takes 5 seconds, previously it took around 50 seconds
```ts
import { setup } from '@nuxt/test-utils/e2e'
import type { TestOptions } from '@nuxt/test-utils/e2e'

[...]

PR #803, opened 2024-03-28
feat: allow to run e2e tests against existing server

tobiasdiez:
Allow to pass the url of an already running server to `setup` and use this as the target instead of building and running a new server. This has two use cases:
1. Run tests against a deployment of your application (i.e. "real" end-to-end tests)
2. During local development, run tests against a already running dev server. This reduces the time to run tests and was wished for in eg https://github.com/nuxt/test-utils/issues/314#issuecomment-1983129269.

Usage:
```ts
setup({ endpoint: process.env.TEST_URL })
```
(or of course just using a hard-coded url)


tobiasdiez:
@danielroe friendly ping. Is there anything a can improve in this PR?

danielroe:
I love this and would like to get this or something like it merged, but just want to make sure we do it in the best way.

danielroe:
@tobiasdiez Is the only use-case here avoiding restarting/rebuilding the server when running tests? (ie. if we can solve that issue, do you still think this is a needed feature?)

adamdehaven:
This PR seems ideal for my use-case as well. We build for the nitro preset `cloudflare-module` which doesn't seem to be compatible out of the box with `@nuxt/test-utils`, meaning I have to override the preset when running tests to `node-server`.

Being able to do a build separately, running the build via `wrangler`, and then pointing the tests to the URL speeds things up exponentially.

The only issue I encounter is that it cannot seem to fetch the manifest:

passing test manifest error

@danielroe I'd be open to other ways of speeding up/reusing a build, but this PR seems promising other than the manifest error above

src/core/types.ts
```diff
@@ -23,6 +23,7 @@ export interface TestOptions {
     launch?: LaunchOptions
   }
   server: boolean
+  endpoint?: string
```
adamdehaven, line 26: suggestion: rename `endpoint` to `host` to sync with Vite (and other) commonly named settings

danielroe:
I think we'll go ahead with this approach. Would you be able to add a test? 🙏

adamdehaven:
This is awesome and will be super useful for us. 

Let me know if the Nuxt docs page needs to be updated and I’m happy to write something up and submit a PR

tobiasdiez:
@danielroe Thanks for merging. I'll try to add tests when I find the time.

@adamdehaven Would be awesome if you could create a PR adding this option to the docs.

adamdehaven:
PR to add the new `host` option, along with a usage example to the Nuxt docs: https://github.com/nuxt/nuxt/pull/28331

Barbapapazes:
This is so cool! Thank you so much, I will give it a try asap.

PR #824, opened 2024-04-22
chore: upgrade `@nuxt/devtools`

danielroe:


danielroe:
@antfu - any ideas why these events are duplicated?

Issue #722, opened 2024-01-18
`renderSuspended` does not provide the values returned from `setup` to the template


obulat:
### Environment

Nuxt project info:                                                                                      

------------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools, modules
- Runtime Modules:  @nuxt/test-utils/module@3.10.0
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-7wbhms?file=test%2Ftest-component.spec.ts

### Describe the bug

`renderSuspended` does not pass the values from `setup` function (not the `<script setup>`) to the `template`. The `mountSuspended` works fine.

### Additional context

In the Openverse Nuxt 3 migration PR, I replaced `render(renderContext, ...args)` with `render(_ctx, ...args)` on this line:

https://github.com/nuxt/test-utils/blob/bc86d23b2cf92aa3c3f19d52944452e4e5c27d63/src/runtime-utils/render.ts#L113

which seems to fix the issue.

joaovitoras:
This is because of defineComponent wrapper
I opened a PR to fix this https://github.com/nuxt/test-utils/issues/722

PR #852, opened 2024-05-20
feat(runtime-utils): expose setup context from `renderSuspended` result

joaovitoras:
After a lot of internal testing in my private project and seeing some similar opened issues about the context returned by setup was not being accessible in the component, I saw that there was a difference in the way how `mountSuspended` and `renderSuspended` were defined.

In `mountSuspended` things always worked, looking at the difference between the two I ended up arriving at the PR https://github.com/danielroe/nuxt-vitest/pull/391 which introduced a fix for this, but only on the mountSuspended function

It would be cool at some point to unify what is possible in these two functions. There is a lot of duplicated logic between  both

Fix: https://github.com/nuxt/test-utils/issues/722

### DIFF
--- a/examples/app-vitest-full/tests/nuxt/render-suspended.spec.ts
+++ b/examples/app-vitest-full/tests/nuxt/render-suspended.spec.ts
@@ -1,11 +1,26 @@
 import { afterEach, describe, expect, it } from 'vitest'
+
 import { renderSuspended } from '@nuxt/test-utils/runtime'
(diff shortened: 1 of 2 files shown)


joaovitoras:
@danielroe This PR is ready! Can you review this or assign someone?
I couldn't find contribution instructions for this repository and I don't know what steps to follow to get this work merged

danielroe (approved):
Very nice - thank you! ❤️

Issue #876, opened 2024-06-28
setProps with array concats newItems and oldItems value


tomoki1207:
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.2
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools, modules
- Runtime Modules:  @nuxt/test-utils/module@3.13.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-ncxs25?file=components%2FTest.spec.ts

### Describe the bug

I called setProps on Component has an Array type prop. However, the resulting prop value is a concatenation of the new array value and the old array value. 
```typescript
const wrapper = await mountSuspended(Test, {
    props: {
      users: [
        { id: 0, name: 'John' },
        { id: 1, name: 'Kate' },
      ],
    },
  });
expect(wrapper.html()).toContain('John'); // props contain John and Kate
await wrapper.setProps({ users: [{ id: 2, name: 'Bob' }] });
expect(wrapper.html()).not.toContain('John');  // !! props contain Bob, John and Kate
```

This issue occurs when using Wrapper from `mountSuspended`, but it does not occur when using @vue/test-utils's `mount`.

What could be the cause of this behavior?

PR #877, opened 2024-06-28
fix(runtime-utils): `setProps` can replace array prop

tomoki1207:
Use custom merge function via `createDefu` at `setProps()` to replace prop that has array type.
Because `defu()` concats array instead of overwrite in default behavior. ([Link](https://github.com/unjs/defu?tab=readme-ov-file#remarks)) 
> Will concat array values (if default property is defined)

Fix: #876


tomoki1207:
@danielroe Could you review this PR or assign someone?

If I have to do any action to contribute for this repository, please tell me about it.

PR #899, opened 2024-07-19
feat(playwright): expose _nuxtHooks fixture setup function

markbrockhoff:
Hi,

I just setup a mock server to mock third party requests made by my nuxt app while running tests with playwright.
In order to redirect those requests to the mock server I overwrote the _nuxtHooks playwright fixture with my own to modify the nuxt config, injecting the url to my mock server into it.

All of this works great but I'd like to keep using the original _nuxtHooks fixture if possible, just modifying the nuxt config passed to it before the app is built.
That's why I extracted it into an exported function so projects can easily modify the nuxt config used for the test app depending on their custom fixtures.

Please let me know what you think :)


markbrockhoff:
Btw: I think the CI pipeline might be broken or flaky as I don't really see a reason why the made change would cause that test to fail on windows ^^'

PR #936, opened 2024-09-09
ctx.nuxt does not exist when build: false in setup

### LINKED ISSUE #918: `ctx.nuxt` does not exist when `build: false` in setup
### Environment

N/A

### Reproduction

N/A

### Describe the bug

setup fails

### Additional context

cc: @tobiasdiez 

Here are the lines

https://github.com/nuxt/test-utils/blob/86d09da23a2d7ba6b0f3dcf2e9ca8c21655a5218/src/core/server.ts#L17-L71

offending line for me was 57, but could easily be any other lines

https://github.com/nuxt/test-utils/commit/c4df41db447afe6101df10fbe763b198b12da86d

is the offending commit 

specifically this change

https://github.com/nuxt/test-utils/commit/c4df41db447afe6101df10fbe763b198b12da86d#diff-84e9c1e393b43cee97ce99909a2db3a09bf9a2435daa2066b01a838e206edec3

### Logs

```shell-script
N/A
```

Al-Rozhkov:
### 🔗 Linked issue

Resolves #918 
resolves https://github.com/nuxt/test-utils/issues/995

### ❓ Type of change

- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)


Boluwatife2904:
Thank You so much for this @Al-Rozhkov. I've been wondering why I've unable to run my tests for a while without any luck.

src/core/server.ts
```diff
@@ -54,7 +54,7 @@ export async function startServer(options: StartServerOptions = {}) {
   }
   else {
     ctx.serverProcess = execa('node', [
-      resolve(ctx.nuxt!.options.nitro.output!.dir!, 'server/index.mjs'),
+      resolve(ctx.options.nuxtConfig.nitro.output!.dir!, 'server/index.mjs'),
```
danielroe, line 57: We should prioritize ctx.nuxt if it exists

Al-Rozhkov: Done. Thank you for quick reply.

danielroe:
Thank you 🙏

Can you explain a bit about why you would do this rather than passing that config to setup?

Al-Rozhkov:
> Thank you 🙏
> 
> Can you explain a bit about why you would do this rather than passing that config to setup?

I wasn't sure how to build Nuxt manually and make setup use exact that build. I would appreciate your advice.

Also not sure why windows tests failed.

Issue #897, opened 2024-07-19
useFakeTimers isn't working for the setInterval.


gsamal:
### Environment

```
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 7.71 GB / 15.38 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
  Browsers:
    Chrome: 112.0.5615.49
  npmPackages:
    vitest: 1.6.0 => 1.6.0

```

### Reproduction

Non-working code with setInterval -
```
  describe('Testing timers', () => {
    let triggered = false;
    beforeAll(() => {
      vi.useFakeTimers();
      setInterval(()=>{
        triggered = true;
      }, 1000)
    });
    afterAll(() => {
      vi.useRealTimers();
    });
    it.only('Should trigger the timer', () => {
      vi.advanceTimersByTime(1000);
      expect(triggered).toEqual(true); // Fails. triggered = false here
    });
  });
```
Working with setTimeout
```
  describe('Testing timers', () => {
    let triggered = false;
    beforeAll(() => {
      vi.useFakeTimers();
      setTimeout(()=>{
        triggered = true;
      }, 1000)
    });
    afterAll(() => {
      vi.useRealTimers();
    });
    it.only('Should trigger the timer', () => {
      vi.advanceTimersByTime(1000);
      expect(triggered).toEqual(true); // Successful.
    });
  });
```

Reproduction - https://stackblitz.com/edit/github-agxb9y-xks33q?file=test%2Fapp.nuxt.spec.ts
Works without nuxt - https://stackblitz.com/edit/vitest-dev-vitest-hr7i1i?file=test%2Ftimer.test.ts

### Describe the bug

`setInterval` isn't triggering with `fakeTimer`, however `setTimeout` does. I have tried `advanceTimersByTime`, `advanceTimersToNextTimer` and `runAllTimers` but doesn't help.

Note - I don't find this issue with the Vitest itself. using `defineVitestConfig` is causing this behaviour.

gsamal:
[...]

PR #1035, opened 2024-12-07
fix: remove `setInterval` from the presets

wattanx:
### 🔗 Linked issue
Fixes: https://github.com/nuxt/test-utils/issues/897

### ❓ Type of change

### 📚 Description
The `setInterval` overridden by Nuxt is auto-imported, so it can no longer be mocked with `useFakeTimers`. 
(`setInterval = window.setInterval`, but the previous state is retained.)

I think updating the documentation with a method like separating workspaces as shown below is one possible approach, but which would be more appropriate?
https://stackblitz.com/edit/github-agxb9y-slbn4d

### DIFF
--- a/examples/app-vitest-full/tests/nuxt/auto-import.spec.ts
+++ b/examples/app-vitest-full/tests/nuxt/auto-import.spec.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 
 describe('auto-imports', () => {
   it('can use core nuxt composables within test file', () => {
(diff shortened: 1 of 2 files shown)


danielroe (approved):
thank you!

should we do this for any of the other polyfills nuxt imports, perhaps?

wattanx:
Yes, that's correct. requestIdleCallback seems to have similar issues as setInterval. 
Adding an option to exclude specific polyfills could be a good approach—for example, something like `imports.polyfills`. What do you think?

danielroe:
maybe!

certainly we can avoid the interval polyfill on client side

Issue #1050, opened 2024-12-25
On Windows, running tests throws error "TypeError: The URL must be of scheme file"


naturalprogrammer:
### Environment

Windows (11 Pro), NodeJS `v22.12.0`

### Reproduction

https://github.com/naturalprogrammer/np-nuxt-test-sample

### Describe the bug

When I run `npm run test` on Windows, I get the following output:

```bash
 FAIL  tests/app.test.ts [ tests/app.test.ts ]
TypeError: The URL must be of scheme file
 ❯ node_modules/@nuxt/test-utils/dirs.js:4:17

 ❯ node_modules/@nuxt/test-utils/dist/e2e.mjs:4:31
```
As a temporary fix, I'd update `node_modules/@nuxt/test-utils/dirs.js` as below:

```ts
import { fileURLToPath } from 'node:url'
import { dirname } from 'path'

// Fix for Windows path resolution
export const distDir = dirname(fileURLToPath(import.meta.url)) + '/dist'
```

### Logs

```shell-script

```

mcrapts:
I have the same issue, but in Linux (not Windows). Happened after upgrading Nuxt from 3.14 to 3.15.

scottbedard:
Duplicate https://github.com/nuxt/test-utils/issues/1038

Issue #1038, opened 2024-12-09
"The URL must be of scheme file"


TheDutchCoder:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-v73onkdp-5xtwbedy?file=vitest.config.ts,globalSetup.ts,test.vitest.e2e.spec.ts

### Describe the bug

Whenever I try to use vitest for e2e testing, the setup function always errors out with a `"The URL must be of scheme file"` error.
The error doesn't tell me anything about what might be wrong, but the runner simply never starts and it has something to do with it expecting a build (even though it's configured to run against localhost instead).

### Additional context

Even when I empty out the vitest config file it doesn't work. I wonder if there's an issue resolving some path internally?

TheDutchCoder:
Turns out the problem is setting `environment: 'nuxt',` in the vitest config.
When I remove that, it works fine.

I'm not sure _why_,.

scottbedard:
I'm also running into this, downgrading to `^3.14.4` works

paambaati:
Same, staying in 3.14.x until this is resolved.

dd-avrillo:
Same here, downgrading to `"@nuxt/test-utils": "3.14.4",` fixed it for me too, thanks @scottbedard. Mac here.

TheDutchCoder:
Didn't work for me unfortunately 😢

HP8585:
I am also facing the same issue rn, nuxt 3.14.4 is fraught with bugs and errors, wearing us down

cybercoder-naj:
I downgraded from `@nuxt/test-utils` from 3.15 to 3.14.4 and I am still facing this issue. :(

FrankoFM:
Downgrading to `3.14.4` is working, everyone having problems with it make sure to not include the `^` in your `package.json`, also deleting `.nuxt` and `node_modules` can help.

HP8585:
even downgrading to 3.14 doesn't work

PR #1054, opened 2025-01-01
fix: use node:path to resolve dist directory

scottbedard:
### 🔗 Linked issue

resolves https://github.com/nuxt/test-utils/issues/1038
resolves https://github.com/nuxt/test-utils/issues/1050

### ❓ Type of change

### 📚 Description

It appears that 56122b7456953cf502b374117427beab1da35f1b caused a regression with 3.15 that can produce the following error. I suspect this is environment-specific. If it were widespread, I'd expect more people to be talking about it.

```
TypeError: The URL must be of scheme file
 ❯ ../node_modules/.pnpm/@nuxt+test-utils@3.15.1_@types+node@22.10.3_@vue+test-utils@2.4.6_happy-dom@15.11.7_magicast@_qhcyoib43mexeh3yv2rdveqtry/node_modules/@nuxt/test-utils/dirs.js:4:17
```

Making this change directly inside `node_modules` works for me, but @danielroe perhaps you can provide more context. If our goal is to simply resolve the `dist` directory, using `node:path` could be a safer way to do it.

### DIFF
--- a/dirs.js
+++ b/dirs.js
@@ -1,7 +1,4 @@
-import path from 'node:path'
+import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
-const __filename = fileURLToPath(import.meta.url)
-const __dirname = path.dirname(__filename)
-
-export const distDir = path.resolve(__dirname, 'dist')
+export const distDir = resolve(dirname(fileURLToPath(import.meta.url)), 'dist')
--- a/dirs.js
+++ b/dirs.js
@@ -1,3 +1,7 @@
+import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 
-export const distDir = fileURLToPath(new URL('./dist', import.meta.url))
+const __filename = fileURLToPath(import.meta.url)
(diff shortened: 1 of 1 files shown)


Aietes:
I think quite a few people have this problem and downgraded test-utils because of it. I can confirm this fix work for me with on a mac.

danielroe:
do you have a reproduction that this fixes? this seems to be a bug in your test runner or node.

scottbedard:
Here is a minimal reproduction

https://github.com/scottbedard/nuxt-test-utils-repro

danielroe:
That's a helpful example. The issue is that the `vitest.config` you have set up is for _Nuxt runtime_ tests (i.e. it sets up a number of things for a browser environment) but you are running _e2e_ tests. For this, I recommend two vitest config files, which target different kinds of tests.

You can see an example in my own website repo: https://github.com/danielroe/roe.dev/blob/main/vitest.config.mts and https://github.com/danielroe/roe.dev/blob/main/vitest.nuxt.config.mts.

danielroe (approved):

Issue #1072, opened 2025-01-14
wrapper.setProps() behaviour for objects is different for mountSuspended vs. mount


sebbayer:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.1
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-o8a9dzsx?file=pages%2F__tests__%2FTestComponent.nuxt.spec.js

### Describe the bug

When I mount a component using `mountSuspended` and then, using `await wrapper.setProps({ ... })`, try to set a prop of type "Object" on that component to `{}` (empty object) or set it to `null`, that prop is not set. See reproduction link.

Also when I call `await wrapper.setProps({ ... })` again after that for the same prop, the contents are **merged** instead of overwritten (as mount does).

This is probably because of how the merging strategy of `defuReplaceArray` works internally in `mountSuspended`.

### Logs

```shell-script

```

PR #1078, opened 2025-01-17
fix: overwrite object props in the same way as `mount`

wattanx:
### 🔗 Linked issue
Fixes: https://github.com/nuxt/test-utils/issues/1072

### ❓ Type of change

### 📚 Description
In `defu`, the first argument takes precedence, so it is not overwritten. 
With this fix, props can now be overwritten just like in `mount`.

### DIFF
--- a/examples/app-vitest-full/components/ExportDefaultComponent.vue
+++ b/examples/app-vitest-full/components/ExportDefaultComponent.vue
@@ -8,6 +8,7 @@
       :key="item"
     >{{ item }}
     </span>
+    <span>myObjProp: {{ JSON.stringify(myObjProp) }}</span>
   </div>
 </template>
 
@@ -30,6 +31,10 @@ export default {
     myArrayProp: {
       type: Array as PropType<string[]>,
       default: () => ([]),
+    },
+    myObjProp: {
+      type: Object as PropType<{ title: string }>,
+      default: () => ({}),
     },
   },
   setup(props) {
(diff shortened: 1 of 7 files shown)


danielroe (approved):
thank you! ❤️

sebbayer:
@wattanx thank you very much for the quick fix, but I think that this solution still does not work for `wrapper.setProps('propName', null)` or setting that prop to `undefined` because of defu.

I would suggest a simple object merge instead of defu, like this:

`return () => h(clonedComponent, { ...props, ...setProps, ...attrs }, slots)`

wattanx:
Isn't the use of wrapper.setProps('prop', null) not an intended use case in the first place?

image

sebbayer:
The only source in the docs I could find was https://vuejs.org/guide/components/props.html#nullable-type

```
export default {
  props: {
    id: {
      type: [String, null],
      required: true
    }
  }
}
```

wattanx:
@sebbayer 

> Isn't the use of wrapper.setProps('prop', null) not an intended use case in the first place?

What I mean is that, in @vue/test-utils, the setProps function does not accept a second argument.
https://stackblitz.com/edit/vitest-tests-browser-examples-yoyouz72

https://test-utils.vuejs.org/api/#setProps
```
setProps(props: Record<string, any>): Promise<void>
```

sebbayer:
Sorry for the confusion, I meant `wrapper.setProps({ propName: null })`.

PR #1108, opened 2025-02-01
fix(runtime-utils): define `$attrs` in component render context

### LINKED ISSUE #1092: `mountSuspended` fails to pass attributes when `inheritAttrs` is false
### Environment

- Operating System: Linux
- Node Version:     v20.15.1
- Nuxt Version:     3.15.2
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Package Manager:  pnpm@8.7.4
- Builder:          -
- User Config:      compatibilityDate, devtools, modules
- Runtime Modules:  @nuxt/test-utils/module@3.15.4
- Build Modules:    -

### Reproduction

https://stackblitz.com/~/github.com/romhml/nuxt-test-repro

### Describe the bug

Hi! There seems to be an issue when using `mountSuspended`   on a component with `defineOptions({ inheritAttrs: false })`   where attributes bound using `v-bind="$attrs"`   are missing.

For instance with this component:

``` vue
<template>
  <div>
    <button v-bind="$attrs"> <slot /> </button>
  </div>
</template>

<script setup lang="ts">
defineOptions({ inheritAttrs: false })
</script>
```

The first test below using `mountSuspended`  will fail because the attribute is missing, whereas the test using ` mount`  from ` @vue/test-utils`  works as expected.

``` ts
import { test, expect } from 'vitest'
import { mount } from '@vue/test-utils'
import { mountSuspended } from '@nuxt/test-utils/runtime'
import ExampleComponent from '~/app/components/ExampleComponent.vue'

test('with mountSuspended', async () => {
  const wrapper = await mountSuspended(ExampleComponent, { attrs: { 'aria-label': 'Aria Label' } })
  expect(wrapper.find('[aria-label="Aria Label"]').exists()).toBe(true)
})


test('with mount', () => {
  const wrapper = mount(ExampleComponent, { attrs: { 'aria-label': 'Aria Label' } })
  expect(wrapper.find('[aria-label="Aria Label"]').exists()).toBe(true)
})
```

### Additional context

_No response_

### Logs

```shell-script

```

romhml:
### 🔗 Linked issue
Resolves #1092 

### ❓ Type of change

### 📚 Description
Fixes an issue where `$attrs`  are undefined when mounting a component using `mountSuspended` .

### DIFF
--- a/examples/app-vitest-full/components/ComponentWithAttrs.vue
+++ b/examples/app-vitest-full/components/ComponentWithAttrs.vue
@@ -1,0 +1,9 @@
+<template>
+  <span v-bind="$attrs" />
+</template>
+
+<script setup lang="ts">
+defineOptions({
+  inheritAttrs: false,
+})
+</script>
--- a/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
+++ b/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
@@ -14,6 +14,7 @@ import ExportDefineComponent from '~/components/ExportDefineComponent.vue'
 import ExportDefaultWithRenderComponent from '~/components/ExportDefaultWithRenderComponent.vue'
 import ExportDefaultReturnsRenderComponent from '~/components/ExportDefaultReturnsRenderComponent.vue'
 import OptionsApiPage from '~/pages/other/options-api.vue'
+import ComponentWithAttrs from '~/components/ComponentWithAttrs.vue'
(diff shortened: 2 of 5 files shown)


danielroe (commented):
can we also apply the same fix to renderSuspended?

romhml:
@danielroe I added a test for `renderSuspended`   but it works without the fix, although attributes are also bound to the wrapper.

romhml:
Done @danielroe! I tweaked the wrapper component a bit because it was inheriting attributes too.

danielroe (approved):
❤️

Issue #1091, opened 2025-01-22
wrapper.element is not reactive for mountSuspended vs. mount


sebbayer:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.2
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools, compatibilityDate
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-w4avpzfe?file=test%2FMyComponent.nuxt.test.js

### Describe the bug

The wrapper returned by `mountSuspended` does not update its `element` property automatically like the wrapper returned by `mount`. See reproduction link. E.g.

```
        const wrapper = await mountSuspended(DynamicContent, { props })
        assert.isTrue(wrapper.element.tagName === 'DIV')
        console.log('before', wrapper.element.tagName)
        await wrapper.setProps({ tag: 'ul' })
        console.log('after', wrapper.element.tagName)
        assert.isTrue(wrapper.element.tagName === 'UL')
```

> before DIV<br>after DIV

If I would use `mount` it would update

> before DIV<br>after UL

This is my workaround:

```
    Object.defineProperty(wrapper, 'element', {
        get () {
            return wrapper.findComponent({ name: 'MountSuspendedComponent' }).element
        },
    })
```

### Logs

```shell-script

```

PR #1110, opened 2025-02-02
fix: add proxy `wrapper.element`

wattanx:
### 🔗 Linked issue
resolves: https://github.com/nuxt/test-utils/issues/1091

### ❓ Type of change

### 📚 Description
`wrapper.element` returned by `mountSuspended` returns the element of `NuxtRoot`, so use a proxy to return the element of the target component.

### DIFF
--- a/src/runtime-utils/mount.ts
+++ b/src/runtime-utils/mount.ts
@@ -96,13 +96,7 @@ export async function mountSuspended<T>(
                     (vm as unknown as AugmentedVueInstance).__setProps = (props: Record<string, unknown>) => {
                       Object.assign(setProps, props)
                     }
-                    vm.props = new Proxy(vm.props, {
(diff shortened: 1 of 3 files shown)


src/runtime-utils/mount.ts
```diff
@@ -214,3 +208,28 @@ function cloneProps(props: Record<string, unknown>) {
...
+        const component = target.findComponent({ name: 'MountSuspendedComponent' })
+        return component[prop]
+      }
+      else {
+        return Reflect.get(target, prop, receiver)
+      }
+    },
+  })
+
+  for (const key of ['props'] as const) {
```
wattanx, line 225: I am using an array so that other methods can also be added.
However, I think isVisible does not need to be proxied. (html and classes might be okay to proxy.

danielroe (approved):
thank you! ❤️

PR #1128, opened 2025-02-19
fix(module): ensure `vi.hoisted` is injected after `vi` import

### LINKED ISSUE #1129: BUG | vitest v3 | `There was an error when mocking a module` when `mockNuxtImport` with existing `vi` import
### Environment

- Operating System: `Linux`
- Node Version:     `v23.5.0`
- Nuxt Version:     `3.15.4`
- CLI Version:      `3.21.1`
- Nitro Version:    `2.10.4`
- Package Manager:  `pnpm@9.15.5`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/nuxt/test-utils/pull/1128

### Describe the bug

when not using vitest globals and importing and using `vi` and `mockNuxtImport` vitest 3 fails with:

```
Error: [vitest] There was an error when mocking a module. If you are using "vi.mock" factory, make sure there are no top level variables inside, since this call is hoisted to top of the file. Read more: https://vitest.dev/api/vi.html#vi-mock
Caused by: TypeError: Cannot read properties of undefined
```

### Additional context

_No response_

### Logs

```shell-script

```

OrbisK:
### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality)

### 📚 Description

This PR fixes #1129 and updates existing vitest example tests to v3.

Some minor linitng issues are fixed as well :)


danielroe:
thank you! i'm aware. there was a breaking change in vitest v3, which the vitest team were aware of before release. i've not yet had time to investigate

OrbisK:
> thank you! i'm aware. there was a breaking change in vitest v3, which the vitest team were aware of before release. i've not yet had time to investigate

FYI:
- works without imports and `globals:  true`

hope it helps

danielroe:
Yes, we're using vitest v4 in the Nuxt repo as we don't require the functionality that's causing the tests to fail here.

OrbisK:
You may already know this, but I dug a little deeper. And I have a feeling that this is caused by the static analysis not picking up the macro.

Form the [docs (warning)](https://vitest.dev/api/vi.html#vi-mock):
> In order to hoist vi.mock, Vitest statically analyzes your files. It indicates that vi that was not directly imported from the vitest package (for example, from some utility file) cannot be used.

OrbisK:
[836ef30](https://github.com/nuxt/test-utils/pull/1128/commits/836ef300c6dbe3e4d9a0f725187eb0aef52d2168)

Has solved it for all tests in this repo. If we need to, maybe we can move the existing import to the top.

(I dont really use monorepos with different versions often. I dont know how to get the lockfile to update to vitest v3 for the example-vitest-3-full 😅)

OrbisK:
> Amazing work! I would be happy simply to upgrade vitest in the monorepo rather than retaining vitest 2 + 3 test suites

Okay, should I update this PR?

danielroe (approved):

test/unit/mock-transform.spec.ts
```diff
@@ -77,10 +77,38 @@ describe('mocking', () => {
...
+      `)
+      expect(code).toMatchInlineSnapshot(`
+        "import {vi} from "vitest";
+        vi.hoisted(() => { 
+                if(!globalThis.__NUXT_VITEST_MOCKS){
+                  vi.stubGlobal("__NUXT_VITEST_MOCKS", {})
+                }
+              });
+
+                import { expect, vi, it } from 'vitest'vi.mock("bob", async (importOriginal) => {
```
danielroe, line 95: should we remove the later import?

OrbisK: we can move this one up instead of creating a new one

OrbisK: I think the first import is just to satisfy vitest analysis. It should not have an actual effect.

OrbisK: This is just guessing:

Maybe the transform is too late for vitest

https://github.com/nuxt/test-utils/blob/main/src/module/plugins/mock.ts#L255

danielroe: maybe just reorder it to beginning of file ....

danielroe: actually, I think I've found the issue - give me a sec

danielroe:
thank you for all your work investigating! ❤️

OrbisK:
> thank you for all your work investigating! ❤️

Thanks for your proper fix.

I tried to go this way too (not as clean tbh), BUT when I did, I encountered 2 things:

### existing vi alias breaks the test

```ts
import { expect, vi as example, it } from 'vitest'
mockNuxtImport('useSomeExport', () => 'bob')
        
it('test', () => {
  const a = example.fn()
})
```

<details><summary>Example test </summary>
<p>

```ts
it('should add `vi` import if existing import is aliased', async () => {
      pluginContext.imports = [{
        name: 'useSomeExport',
        from: 'bob',
      }]
      const code = await getResult(`
        import { expect, vi as iv, it } from 'vitest'
        mockNuxtImport('useSomeExport', () => 'bob')
        
        it('test', () => {
          const a = iv.fn()
        })
      `)
      expect(code).toMatchInlineSnapshot(`
        "
                import { expect, vi as iv, it } from 'vitest'
        vi.hoisted(() => { 
                if(!globalThis.__NUXT_VITEST_MOCKS){
                  vi.stubGlobal("__NUXT_VITEST_MOCKS", {})
                }
              });

        vi.mock("bob", async (importOriginal) => {
          const mocks = globalThis.__NUXT_VITEST_MOCKS
          if (!mocks["bob"]) {
            mocks["bob"] = { ...await importOriginal("bob") }
          }
          mocks["bob"]["useSomeExport"] = await (() => 'bob')();
          return mocks["bob"] 
        });

                
                
                it('test', () => {
                  const a = iv.fn()
                })
              
         import "bob";"
      `)
    })
```

</p>
</details> 

### The previously generated code was valid test code.

The code before the changes from this PR actually worked when pasted into a `*.spec.ts`.


-------------

I hope you can share your thoughts on this 😅

danielroe:
i’m sure we could resolve.

PR #1135, opened 2025-02-21
fix(runtime-utils): ensure event listeners props are called

### LINKED ISSUE #572: Emitted events doesn't work with defineModel
In `v0.8.1` emitted events were fixed in mountSuspended, but they still doesn't work with `defineModel`.

Added some tests for it in nuxt/nuxt-vitest#193.

n-rowe:
### 🔗 Linked issue

Resolved #572

### ❓ Type of change

- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

With the change to resolve #572 the internal emit function for the component was changed to use the root component so that emits can be picked up by vitest.

This had the side-effect of breaking user-passed event listeners as props to the component:
```js
const component = await mountSuspended(MyComponent, {
  props: {
    'onUpdate:modelValue': async e => component.setProps({ modelValue: e }),
  },
})
```

To resolve this I changed the internal component emit to call both the original emit and the root component emit.
I've also added a new unit test for this case.

### DIFF
--- a/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
+++ b/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
@@ -123,6 +123,19 @@ describe('mountSuspended', () => {
     expect(component.emitted()).toHaveProperty('update:modelValue')
   })
 
+  it('can pass onUpdate event to components using defineModel', async () => {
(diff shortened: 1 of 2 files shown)


danielroe (approved):

danielroe:
thank you!

PR #1221, opened 2025-03-29
fix: rename `setup` hook to `beforeAll`

BobbieGoede:
### ❓ Type of change

- [ ] ?  ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This may be a breaking change, I'm not sure if these types/functions are fully internal or if they are exposed for users 🤔

This PR renames `setup` to `beforeAll` to better communicate when and how it is used (internally only?), to be consistent with the other hook functions. The `setup` hook is marked deprecated and should still work, but should be removed in a future major bump.

### DIFF
--- a/src/core/setup/index.ts
+++ b/src/core/setup/index.ts
@@ -67,6 +67,7 @@ export function createTest(options: Partial<TestOptions>): TestHooks {
     afterEach,
     afterAll,
     beforeAll,
+    setup: beforeAll,
     ctx,
   }
 }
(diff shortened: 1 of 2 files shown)


BobbieGoede:
Ah seeing the failing test, this is likely a breaking change, perhaps this is something to consider in a major release someday, otherwise feel free to close

src/core/setup/index.ts
```diff
@@ -66,7 +66,7 @@ export function createTest(options: Partial<TestOptions>): TestHooks {
     beforeEach,
     afterEach,
     afterAll,
-    setup,
+    beforeAll,
```
danielroe, line 69: to avoid a breaking change we could also return setup but mark it as deprecated

BobbieGoede: I have added `setup` to the return and marked it deprecated

danielroe (approved):

PR #1311, opened 2025-06-02
fix: Include tests without `.nuxt.` extension

### LINKED ISSUE #1296: Update from 3.18 to 3.19 changed default configuration, causing tests without .nuxt. to not be included
### Environment

- Operating System: `Darwin`
- Node Version:     `v22.12.0`
- Nuxt Version:     `3.17.3`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `bun@1.2.14`
- Builder:          `-`
- User Config:      `modules`, `typescript`, `alias`, `compatibilityDate`, `css`, `devtools`, `experimental`, `future`, `rootDir`, `vite`, `hanko`, `i18n`
- Runtime Modules:  `~~/modules/startup-check`, `@nuxt/fonts@0.11.4`, `@nuxt/icon@1.13.0`, `@nuxt/image@1.10.0`, `@nuxt/test-utils@3.19.0`, `@nuxt/test-utils/module@3.19.0`, `@nuxt/ui@3.1.2`, `@nuxtjs/i18n@9.5.4`, `@nuxtjs/hanko@1.0.0`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-d5xpujkj?file=test%2Fbug.test.ts

### Describe the bug

I got some trouble when I updated from 3.18 to 3.19

After the update only `.nuxt.test.ts` tests are run by vitest.

To restore the previous behaviour (also running tests unrelated to nuxt), I now need to set environment "nuxt", which feels really unintiutive.
```TypeScript
export default defineVitestConfig({
  test: {
    environment: "nuxt",
  },
})
```

### Additional context

_No response_

### Logs

```shell-script

```

lutejka:
Include tests without `.nuxt.` extension, by creating a workspace for the default environment

### 🔗 Linked issue

Resolves #1296

### ❓ Type of change

### 📚 Description

Moving environmentMatchGlobs to define a workspace for the nuxt environment resulted in Vitest only running tests within that specific nuxt workspace, excluding tests from the parent/default configuration (from which the workspace inherits). Creating an additional workspace for the default, non-Nuxt environment ensured all tests were included again.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,7 @@
 import process from 'node:process'
 import type { Nuxt, NuxtConfig } from '@nuxt/schema'
 import type { UserWorkspaceConfig, InlineConfig as VitestConfig } from 'vitest/node'
-import { defineConfig } from 'vitest/config'
+import { defineConfig, defaultExclude } from 'vitest/config'
 import type { TestProjectInlineConfiguration } from 'vitest/config'
 import { setupDotenv } from 'c12'
 import type { DotenvOptions } from 'c12'
@@ -261,6 +261,18 @@ export function defineVitestConfig(config: ViteUserConfig & { test?: VitestConfi
           include: [
             '**/*.nuxt.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}',
             '{test,tests}/nuxt/**.*',
+          ],
+        },
+      })
+      resolvedConfig.test.workspace.push({
+        extends: true,
+        test: {
+          name: defaultEnvironment,
+          environment: defaultEnvironment,
+          exclude: [
+            ...defaultExclude,
+            './**/*.nuxt.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}',
+            './{test,tests}/nuxt/**.*',
           ],
         },
       })


src/config.ts
```diff
@@ -2,7 +2,7 @@ import process from 'node:process'
 import type { Nuxt, NuxtConfig } from '@nuxt/schema'
 import type { UserWorkspaceConfig, InlineConfig as VitestConfig } from 'vitest/node'
 // this is deliberately the vite config function so the module runs if vitest is not installed
-import { defineConfig } from 'vite'
+import { defineConfig, defaultExclude } from 'vitest/config'
```
danielroe, line 5: see comment above it - we should not import from `vitest` in this file

danielroe (commented):
this looks great! might it be possible to add a simple fixture to the `examples/` directory so we avoid this regression happening again?

src/config.ts
lutejka: So what should we do here? Should I avoid importing `defaultExclude` all together. Or is it fine to to still import `defaultExclude` in an extra import from `vitest/config`? 

If we want to avoid it, the only option I see is copying the exclude list into this file....

danielroe: we could use a dynamic import and catch to return [], or inline the exclusions

the reason is that this module may be installed by people without vitest, even though this particular code requires it

lutejka:
> this looks great! might it be possible to add a simple fixture to the `examples/` directory so we avoid this regression happening again?

Good idea! Adding the test showed, that the fix will not apply when the tests run as part of the test server. (`pnpm test:dev`) The reason for this is probably the same why we need to add `environmentMatchGlobs`...


Unfortunately, I could not get it running, even when adding the workspaces for this case. Any idea why?

danielroe:
I think we could merge this fix as-is (as it is a very significant issue) and add a todo to fix tests run from the dev server in future. Anthony & I already discussed potentially disabling that feature in future.

danielroe (approved):

danielroe:
thank you ❤️

PR #1328, opened 2025-06-17
fix(runtime-utils): event duplication

### LINKED ISSUE #1176: Component mounted with mountSuspended emits events twice after setProps was called
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.4
- CLI Version:      3.22.4
- Nitro Version:    2.11.3
- Package Manager:  npm@10.2.3
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-4czkz34t?file=components%2FMyComponent.vue,test%2FMyComponent.test.js

### Describe the bug

After using `setProps` in a component mounted with `mountSuspended`, using `$emit` inside a method will emit the event twice, which can be checked with `wrapper.emitted()`. This behaviour does not exist when using `mount()`. Please see reproduction link.

A workaround is to use the `$emit` directly in the `@click` attribute, like this:

```
<button @click="$emit('test', 'a')">test</button>
```

But this is not sufficient if a method needs to be called that does other things as well besides emitting.

### Additional context

_No response_

### Logs

```shell-script

```

Thomaash:
### 🔗 Linked issue

resolves #1176

### ❓ Type of change

### 📚 Description

The emit function was intercepted again on every rerender of the component. This caused the events in `wrapper.emitted()` to be duplicated more and more times as the test went on. This changes the logic to check if the emit function was already intercepted and if so, doesn't intercept it again.

### DIFF
--- a/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
+++ b/examples/app-vitest-full/tests/nuxt/mount-suspended.spec.ts
@@ -123,6 +123,50 @@ describe('mountSuspended', () => {
     expect(component.emitted()).toHaveProperty('update:modelValue')
   })
 
+  it('can receive emitted events from components mounted within nuxt suspense using defineModel after prop changes and multiple interactions', async () => {
+    const component = await mountSuspended(WrapperTests)
+
+    component.find('button#changeModelValue').trigger('click')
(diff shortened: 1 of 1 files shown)


danielroe (approved):
thank you ❤

PR #1353, opened 2025-07-08
fix(runtime-utils): capture events emitted before the first render

### LINKED ISSUE #1332: `mountSuspended` does not catch event emitted in immediate watcher with options API
### Environment

- Operating System: Linux
- Node Version:     v22.16.0
- Nuxt Version:     3.17.5
- CLI Version:      3.25.1
- Nitro Version:    2.11.12
- Package Manager:  npm@11.4.1
- Builder:          -
- User Config:      compatibilityDate, devtools, future, srcDir, serverDir, dir, appId, app, modules, ssr, css, esbuild, vite, typescript, vue, postcss, phosphor, alias, components, imports, sentry, sourcemap, nitro, hooks
- Runtime Modules:  @nuxt/test-utils/module@3.19.1, @pinia/nuxt@0.11.1, @sentry/nuxt/module@9.29.0, @vee-validate/nuxt@4.15.1, @vueuse/nuxt@13.3.0, nuxt-phosphor-icons@1.3.0
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/devbox/p2qm2h

- run `npm run test`
- see that the options API test fails as no events are captured when using `mountSuspended`
- see that it works correctly when using `mount` from @vue/test-utils

### Describe the bug

Initial events are not captured when using options API inside a component.

### Additional context

_No response_

### Logs

```shell-script

```

Thomaash:
### 🔗 Linked issue

resolves #1332

### ❓ Type of change

### 📚 Description

Immediate watches are executed before the first render but since emit was intercepted in render function, their events were lost. Intercepting emit in setup solves this issue.


src/runtime-utils/mount.ts
```diff
@@ -178,7 +202,7 @@ export async function mountSuspended<T>(
                             return render.call(this, renderContext, ...args)
                           }
                           : undefined,
-                        setup: setup ? (props: Record<string, unknown>) => wrappedSetup(props, setupContext) : undefined,
+                        setup: (props: Record<string, unknown>) => wrappedSetup(props, setupContext),
```
danielroe, line 205: was there a reason you changed this line?

src/runtime-utils/mount.ts
```diff
@@ -107,28 +147,12 @@ export async function mountSuspended<T>(
...
 
-                      let interceptedEmit: ((event: string, ...args: unknown[]) => void) | null = null
-
                       // Proxy top-level setup/render context so test wrapper resolves child component
                       const clonedComponent = {
                         name: 'MountSuspendedComponent',
                         ...component,
                         render: render
                           ? function (this: unknown, _ctx: Record<string, unknown>, ...args: unknown[]) {
-                            // When using defineModel, getCurrentInstance().emit is executed internally. it needs to override.
```
danielroe, line 118: do we need to reintercept the emit here if currentInstance (above) was null?

src/runtime-utils/mount.ts
```diff
@@ -59,15 +59,63 @@ export async function mountSuspended<T>(
...
+   * Intercept emit for assertions in populate wrapper emitted.
+   */
+  function interceptEmitOnCurrentInstance(): void {
+    const currentInstance = getCurrentInstance()
+    if (currentInstance == null) {
+      return
+    }
+
+    currentInstance.emit = getInterceptedEmitFunction(currentInstance.emit)
+  }
```
danielroe, line 111: I think we also need to apply this fix in `render.ts` (and maybe we could consider abstracting the code in a future PR so we don't just have two copies of the same logic)

src/runtime-utils/mount.ts
```diff
@@ -178,7 +212,7 @@ export async function mountSuspended<T>(
                             return render.call(this, renderContext, ...args)
                           }
                           : undefined,
-                        setup: setup ? (props: Record<string, unknown>) => wrappedSetup(props, setupContext) : undefined,
+                        setup: (props: Record<string, unknown>) => wrappedSetup(props, setupContext),
```
danielroe, line 215:
```suggestion
                        setup: setup ? (props: Record<string, unknown>) => wrappedSetup(props, setupContext) : undefined,
```

danielroe:
thank you ❤️

PR #1389, opened 2025-08-19
feat(runtime): add `scoped` option to cleanup components

### LINKED ISSUE #517: Clearing state before/after each test
I am writing a few tests that are changing some internal `useState` variables of a composable.
However, I noticed that the state of that composable is kept between tests.

This breaks a primary principle of unit testing: 
1. tests should be isolated.
2. tests should work in any order they are run (dependent on 1.)

Even applying 
```ts
beforeEach(() => {
    clearNuxtState();
}
```
the composable still keeps the state.

### LINKED ISSUE #1367: Mounted components do not automatically unmount once the `it` statement finishes
### Environment

I got this on my project, but was able to reproduce it on the stackblitz starter. Here's the info from there:

```
Working directory: /home/projects/github-rg6uhsqu                                                                                                                                         nuxi 15:08:38
Nuxt project info:                                                                                                                                                                        nuxi 15:08:38

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     4.0.1
- CLI Version:      3.26.4
- Nitro Version:    2.12.4
- Package Manager:  npm@10.8.2
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

Reproduction: https://stackblitz.com/edit/github-rg6uhsqu?file=components%2Ftests%2FMyComponent.spec.ts

1. Run `npm run test`
2. See that the `console.log` from the watcher triggers both from Component 1 and Component 2; even though the state is only changed in the second test
3. Unmounting the first wrapper within it's `it` block fixes the issue

### Describe the bug

It seems that wrappers do not unmount automatically after their it statements. I'm not sure if this is expected behaviour, but I wouldn't expect it and I'm worried that it can cause some data-contamination between tests and cause memory leaks?

### Additional context

_No response_

### Logs

```shell-script

```

MikeBellika:
### 🔗 Linked issue

resolves #517
resolves #1367

### ❓ Type of change

### 📚 Description

This PR adds a `scoped` option to `mountSuspended`. When this option is enabled, state isn't shared between components. This is achieved by using an [effectScope](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md) and a "global" cleanup function. This is similar to what is done in https://github.com/nuxt/test-utils/blob/main/src/runtime-utils/render.ts

The watcher cleanup tests are adapted from #1367. The `useCounter` tests are adapted from https://github.com/MikeBellika/component-not-unmounted-test-utils/blob/main/test/unit/repro.test.ts

### DIFF
--- a/src/runtime-utils/mount.ts
+++ b/src/runtime-utils/mount.ts
@@ -1,7 +1,6 @@
 import { mount } from '@vue/test-utils'
 import type { ComponentMountingOptions } from '@vue/test-utils'
-import { Suspense, h, isReadonly, nextTick, reactive, unref, getCurrentInstance, effectScope } from 'vue'
-import { Suspense, h, isReadonly, nextTick, reactive, unref, getCurrentInstance, isRef } from 'vue'
+import { Suspense, h, isReadonly, nextTick, reactive, unref, getCurrentInstance, effectScope, isRef } from 'vue'
 import type { ComponentInternalInstance, DefineComponent, SetupContext } from 'vue'
 import { defu } from 'defu'
 import type { RouteLocationRaw } from 'vue-router'


MikeBellika:
> The difference is that `render` only expects a single component to render, but you might mount more than one component in a test.
> 
> Additionally, some people mount a component and then have several assertions (or 'it' tests) about it.
> 
> I think it's better to require people to unmount components manually, which is also (afaik) what you have to do if you are using `@vue/test-utils`.

I think it's unintuitive that state is shared between tests and wouldn't expect the order of calling tests to influence the result. If people do multiple assertions without mounting the component again I think they're writing their tests wrong.

danielroe:
A test library like this should be unopinionated about how people do their mounting of components.

I think we should follow the example of `@vue/test-utils` when it comes to `mountSuspended`...

Some code that would be broken:

```ts
it('implicitly unmounts comp1, unexpectedly', async () => {
  const comp1 = await mountSuspended(/* */)
  const comp2 = await mountSuspended(/* */)
})
```

MikeBellika:
# Investigation
Just checked in vue: https://github.com/MikeBellika/vue-test-mount-repro
The tests pass but the console logs in [HelloWorld.vue](https://github.com/MikeBellika/vue-test-mount-repro/blob/main/src/components/HelloWorld.vue) aren't reached. So I guess the components aren't unmounted between tests and effectScope isn't used either.

Just tried to run this test with my change:
```typescript
  it('can mount twice', async () => {
    useCounterMock.mockRestore()
    const comp1 = await mountSuspended(GenericStateComponent)
    console.log('comp1 mounted')
    const comp2 = await mountSuspended(GenericStateComponent)
    console.log('comp2 mounted')
    expect(comp1.text()).toMatchInlineSnapshot('"Zero or negative count"')
    expect(comp2.text()).toMatchInlineSnapshot('"Zero or negative count"')
  })
```
I also updated `GenericStateComponent`
```vue
<script setup lang="ts">
const { isPositive } = useCounter()
const { data } = useLazyAsyncData(async () => {
  return isPositive() ? 'positive' : 'zero-or-negative'
})

onUnmounted(() => {
  // not called
  console.log('unmounted')
})

onScopeDispose(() => {
  console.log('disposed')
})
</script>

<template>
  <div v-if="data === 'positive'">
    Positive count
  </div>
  <div v-else>
    Zero or negative count
  </div>
</template>
```

The output is:
```
stdout | tests/nuxt/mount-suspended.spec.ts > composable state isolation > can mount twice
disposed

stdout | tests/nuxt/mount-suspended.spec.ts > composable state isolation > can mount twice
comp1 mounted
disposed

stdout | tests/nuxt/mount-suspended.spec.ts > composable state isolation > can mount twice
comp2 mounted
```

So the components aren't unmounted within the test but the scopes do seem to be disposed. I don't know what effect this will have.

# My opinion
As a user of nuxt test-utils the behaviour of different invocations of `mountSuspended` reusing the same state is unexpected. I don't think I'm alone in this and IMO #517 and #1367 stem from this issue.

My mental model of `mountSuspended` is that it mounts a component in isolation, with nothing shared between other tests. From my tests it seems like this is how it works in vue but I could easily be mistaken.
If this is the wrong way to think about it, I think a section about test isolation should be added to the documentation. Maybe it could have some tips about using `component.unmount()` and `clearNuxtData()` after each test.

Alternatively what about an `isolate` option on `mountSuspended`?

danielroe:
I'd be very happy to support an option (maybe `dispose`?)

however, isolated is a bit misleading as there's a single `nuxtApp` instance that persists between tests, and things like state (via `useState`) doesn't reset. we already call this out IIRC but maybe we can make it clearer

MikeBellika:
What about `scoped` as a name for the option? 

---

I'm not too strong on Nuxt internals but I think there are two related issues that I have gotten a bit mixed up.
1. State like `useState` is shared between tests. This is mentioned in the docs but could be made more explicit
2. [this](https://github.com/MikeBellika/component-not-unmounted-test-utils/blob/main/test/unit/repro.test.ts) test fails. I think it's because the function in `useLazyAsyncData` is reused between tests. To me this feels like a foot gun.

What do you think about:
1. Updating [docs](https://nuxt.com/docs/4.x/getting-started/testing) to include a section about what state is shared between tests and how to ensure test isolation
2. Besides adding an option on `mountSuspended`, adding an option to create a whole new `nuxtApp` instance for each test.

I can do 1. but I wouldn't know where to start with 2.

Regarding 1. the docs do mention that a global nuxt instance will be used but I think it could be made more clear:
> When you run your tests within the Nuxt environment, they will be running in a [happy-dom](https://github.com/capricorn86/happy-dom) or [jsdom](https://github.com/jsdom/jsdom) environment. Before your tests run, a global Nuxt app will be initialized (including, for example, running any plugins or code you've defined in your app.vue). This means you should take particular care not to mutate the global state in your tests (or, if you need to, to reset it afterwards).

Source: https://nuxt.com/docs/4.x/getting-started/testing#running-tests

MikeBellika:
@danielroe I've updated the PR to add a `scoped` option

PR #1396, opened 2025-08-29
test: adding example tests for NuxtUI

JessicaSachs:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

A reference Nuxt UI example would have been helpful for me to validate that CSS and Tailwind and Nuxt instance-dependent components would work.

This is slightly redundant with the `nuxt-vitest-browser` example, but because we have a `nuxt-vitest-content` example, I think it's worth adding.

### DIFF
--- a/examples/nuxt-ui/.gitignore
+++ b/examples/nuxt-ui/.gitignore
@@ -1,0 +1,1 @@
+.data/content/*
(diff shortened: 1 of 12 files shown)


danielroe (approved):

danielroe:
thank you 🙏

PR #1475, opened 2025-10-29
feat(runtime-utils): support once option in registerEndpoint

wattanx:
### ❓ Type of change

- [x] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description
Add `once` option to `registerEndpoint` function that automatically removes the handler after the first matching request,
  similar to msw's `once` behavior.
With the "once" option, you can configure it to return different responses for the same endpoint.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ env:
   # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
   NODE_OPTIONS: --max-old-space-size=6144
 
-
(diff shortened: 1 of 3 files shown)


src/runtime-utils/mock.ts
```diff
@@ -18,30 +18,39 @@ import type {
 type Awaitable<T> = T | Promise<T>
 type OptionalFunction<T> = T | (() => Awaitable<T>)
 
+const endpointRegistry: Record<string, Array<{ handler: EventHandler, method?: HTTPMethod, once?: boolean }>> = {}
```
wattanx: I moved it because it wasn’t a comment for registerEndpoint.

PR #1501, opened 2025-12-05
fix(module): ensure imports are collected before mock transform runs

danielroe:
### 📚 Description

this aims to address the cause of some flaky tests in nuxt/test-utils itself - indicating we might have a race condition when running the transform for imports


danielroe:
~> https://github.com/nuxt/test-utils/commit/79f1e14d5079101ee19b0fc3f68ed55e73b5c662

PR #1512, opened 2025-12-11
perf(module): use filter for NuxtRootStubPlugin

danielroe:
### ❓ Type of change

### 📚 Description

we should also entirely avoid injecting the plugin when building!


src/module/plugins/entry.ts
```diff
@@ -1,6 +1,6 @@
 import { readFileSync } from 'node:fs'
 import { dirname, extname, join } from 'pathe'
-import { createUnplugin } from 'unplugin'
+import type { Plugin } from 'vite'
```
TheAlexLichter: unplugin also has hook filter support, so you should be able to stick with it (if wanted)

TheAlexLichter (approved):

PR #1522, opened 2025-12-20
feat(runtime-utils): unify logic of mount + render helpers

yamachi4416:
### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality)

### 📚 Description

Unified the underlying logic of the `mountSuspended` and `renderSuspended`.
This also fixes the missing setup logic for vueApp.context.components in renderSuspended.


danielroe:
thank you!

one thing is that there's a lot of overlap between the two implementations

i keep meaning to refactor out the shared pieces so bug fixes in one propagate to the other

yamachi4416:
thank you for the review as always. 
You're absolutely right. I'm going to refactor it to pull out the shared logic.

yamachi4416:
thank you for the advice.
I’ve refactored the internal logic of mountSuspended and renderSuspended to share common code.
I would appreciate it if you could take a look.

PR #1708, opened 2026-05-20
fix(config): nested browser-mode Nuxt Vitest projects with `viteEnvironmentApi`

peterbud:
### 🔗 Linked issue

Resolves #1689

### 📚 Description

This PR fixes when `defineVitestProject(...)` is used inside `test.projects` for browser-mode tests on Vitest 4 and the Nuxt app enables `experimental.viteEnvironmentApi`.

Before this change, the same Nuxt browser project behaved differently depending on how it was consumed:

- nested under `test.projects`: browser session timed out before any test ran
- exported as the top-level config: tests passed normally

The fix keeps the browser project on the correct Vite client transform path so nested browser projects behave the same as top-level browser projects.

## Problem

The failure was reproducible in a minimal Nuxt fixture with only:

- `defineVitestProject(...)`
- `browser.enabled: true`
- `experimental.viteEnvironmentApi: true`

The minimal failing shape was:

```ts
import { fileURLToPath } from 'node:url'
import { defineVitestProject } from '@nuxt/test-utils/config'
import { playwright } from '@vitest/browser-playwright'
import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
    projects: [
      await defineVitestProject({
        test: {
          name: 'nuxt-browser',
          environment: 'nuxt',
          include: ['app/components/**/*.test.ts'],
          environmentOptions: {
            nuxt: {
              rootDir: fileURLToPath(new URL('.', import.meta.url)),
            },
          },
          browser: {
            enabled: true,
            provider: playwright(),
            instances: [{ browser: 'chromium' }],
            headless: true,
          },
        },
      }),
    ],
  },
})
```

The matching top-level export using the same project config passed.

## Root cause

The actual break happened at the Vite environment boundary used by the Vitest browser runner:

- in the failing nested case, the resolved Vite `client` environment ended up with `dev.moduleRunnerTransform = true`
- that caused Vitest browser runner assets such as `__vitest_browser__/orchestrator-*.js` to be served with SSR-style `__vite_ssr_import__` rewrites
- those assets execute in the browser, where `__vite_ssr_import__` does not exist
- the orchestrator page crashed during bootstrap, so the browser websocket never connected
- Vitest then surfaced the symptom as `Failed to connect to the browser session ... within the timeout`

In the passing cases, the same assets were served as normal browser ESM imports.

## Fix in this PR

When browser mode is enabled, `@nuxt/test-utils` now adds a Vite `configEnvironment` hook for the `client` environment and forces: `dev.moduleRunnerTransform = false`

That keeps the browser runner assets on the correct client-side transform path even when the Nuxt project is returned inline through `test.projects`.


yamachi4416:
Thank you!
This fix seems to work fine
- before: https://stackblitz.com/edit/nuxt-test-utils-pull-1708-before?file=package.json
- after: https://stackblitz.com/edit/nuxt-test-utils-pull-1708-after?file=package.json

It seems like the same fix is already in vitest's main branch: https://github.com/vitest-dev/vitest/blob/22d353a80c230dbc7c69a22b390070f7aa9afabe/packages/vitest/src/node/plugins/runnerTransform.ts#L77-L79

I think this fix is necessary now. (hopefully, it will be backported to v4)

danielroe (approved):

PR #1720, opened 2026-06-09
build: migrate to tsdown

danielroe:
### 📚 Description

updates build config following pattern in https://github.com/nuxt/nuxt/pull/35179

- improves type checking
- removes need for stub mode


src/module.ts
```diff
@@ -55,6 +55,14 @@ export default defineNuxtModule<NuxtVitestOptions>({
...
     nuxt.hook('prepare:types', (ctx) => {
       ctx.references.push({ types: 'vitest/import-meta' })
+      // TODO: drop once Nuxt enables this by default for source-pointing exports.
+      // `@nuxt/test-utils` exports resolve to `.ts` source at dev time, so consumer
+      // typechecks need this flag to follow them.
+      for (const tsConfig of [ctx.tsConfig, ctx.nodeTsConfig, ctx.sharedTsConfig]) {
+        if (!tsConfig) continue
+        tsConfig.compilerOptions ||= {}
+        tsConfig.compilerOptions.allowImportingTsExtensions = true
+      }
```
danielroe, line 65: this is coming in nuxt v4.5 so we wouldn't ship this code, it's just temporary for testing

yamachi4416 (approved):

PR #1725, opened 2026-06-16
feat(browser): add render helper for vitest browser mode

yamachi4416:
### 📚 Description

Added a render helper with Suspense support for Vitest Browser Mode on Nuxt.
Initially, I considered wrapping `vitest-browser-vue`.
However, I decided to implement it on our side instead, as it requires some adjustments and gives us more flexibility.

I would appreciate your feedback.


danielroe (approved):
@yamachi4416 this is amazing! I updated it with a few changes (some significant, like [this one](https://github.com/nuxt/test-utils/pull/1725/changes/9a727785720d94ecc0ea4edfb72b9349036f6218)) - are you happy with the latest state?

yamachi4416:
@danielroe 
​thank you so much as always.
I'm happy! I think [this one](https://github.com/nuxt/test-utils/pull/1725/changes/9a727785720d94ecc0ea4edfb72b9349036f6218) make testing much more natural and customizable.

PR #1744, opened 2026-07-20
chore(deps): pin nitro and h3-next version

yamachi4416:
### 📚 Description

Fix the CI test failure in the renovate PR for h3-next.
https://github.com/nuxt/test-utils/pull/1742

### DIFF
--- a/examples/nitro-v3/package.json
+++ b/examples/nitro-v3/package.json
@@ -11,8 +11,8 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "h3": "^2.0.1-rc.25",
-    "nitro": "^3.0.260610-beta",
+    "h3": "2.0.1-rc.22",
+    "nitro": "3.0.260610-beta",
     "nuxt": "npm:nuxt-nightly@5x"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


examples/nitro-v3/package.json
```diff
@@ -11,8 +11,8 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "h3": "^2.0.1-rc.22",
-    "nitro": "^3.0.260311-beta",
+    "h3": "^2.0.1-rc.25",
```
danielroe: nitro currently pins h3@2.0.1-rc.22 so we probably need to pin both h3 + nitro here

yamachi4416: @danielroe 
Thank you!
~~Should we also add the renovate configuration in the same way?~~
I’ve applied the changes based on the PR below. Could you please take a look?
https://github.com/nuxt/nuxt/pull/35555/changes#diff-7b5c8955fc544a11b4b74eddb4115f9cc51c9cf162dbffa60d37eeed82a55a57

danielroe (approved):

PR #1759, opened 2026-08-01
chore: align engines.node to nuxt

yamachi4416:
### 📚 Description

Remove EOL Node 20 from `engines.node` and align with Nuxt


danielroe:
totally fine with this, or we could wait until we are blocked by a dependency or feature that needs a newer node version

yamachi4416:
Thank you for the review. makes sense to wait until we hit a blocker.

Issue #1388, opened 2025-08-19
Error when loading browser-entry.mjs: Syntax error, "qs" does not provide an export named "default"


andreaswolf:
### Environment

- Operating System: Linux
- Node Version:     v20.19.4
- Nuxt Version:     3.18.1
- CLI Version:      3.27.0
- Nitro Version:    2.12.4
- Package Manager:  yarn@4.2.2
- Builder:          -
- User Config:      devtools, modules, ui, colorMode, css, formkit, pinia, components, ssr, typo3
- Runtime Modules:  @nuxt/ui@3.3.0, @pinia/nuxt@0.11.2, @t3headless/nuxt-typo3@2.2.1, @formkit/nuxt@1.6.9, @nuxtjs/storybook@9.0.0, @nuxt/test-utils/module@3.19.2, nuxt-viewport@2.3.1
- Build Modules:    -

### Reproduction

See https://github.com/andreaswolf/nuxt-issue-repro

Run `yarn run vitest --project=e2e` for a working test, `yarn run vitest --project=e2e-nuxt` will not work.

### Describe the bug

Tests run fine without Nuxt/defineVitestProject. As soon as I use defineVitestProject(), there are two errors and the tests are not run anymore:

1. Error: Failed to import test file /home/aw/Kunden/fes/fes-cp-v11/frontend-v2/node_modules/@nuxt/test-utils/dist/runtime/browser-entry.mjs
Caused by: SyntaxError: The requested module '/node_modules/qs/lib/index.js?v=c5ced9fb' does not provide an export named 'default'

2. TypeError: Cannot set properties of undefined (setting 'PROD')

The second issue (very likely) comes from setupEnv in vitest's setup-common.js. I can also open a separate issue for this.

### Logs

```shell-script

```

yamachi4416:
It works by adding `qs` to `vite.optimizeDeps.include` in nuxt.config.ts or to the project's `optimizeDeps.include` in vitest.config.ts.
```ts
export default defineNuxtConfig({
  vite: {
    optimizeDeps: {
      include: ['qs'],
    },
  },
})
```
https://stackblitz.com/edit/nuxt-test-utils-issues-1388?file=vitest.config.ts,nuxt.config.ts

PR #1775, opened 2026-08-08
chore: avoid missing file extension warning in test/examples

yamachi4416:
### 📚 Description

~~Add test for cjs module adding to `vite.optimizeDeps.include` in nuxt.config.ts, and post a comment to a solution to the issue:  #1388~~

avoid missing file extension warning
> import "./app5/bare.vitest.config" without a file extension (vitest.config.ts:7:14). Add the file extension

ignore vitest test results in .gitignore


danielroe (approved):
do we really need a test fixture for this?

yamachi4416:
@danielroe 
Thank you for review.
for test-utils, we might not need a test case for this.
I’ll scope down the PR to just fixing the missing file extension warning and updating .gitignore.

danielroe (approved):

PR #1797, opened 2026-08-31
test: make sure tests are collected at runtime, not statically

sheremet-va:
### 📚 Description

In the latest Vitest 5 rc we changed the default for `vitest.collect` to statically collect tests instead of running them at runtime. The option exists since 4.1, this just makes sure the behaviour in the Nuxt test is consistent between the majors.

See https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/33385971363/job/99468551633


sheremet-va:
> We've flagged this as a potential contribution without a human behind it. 

giphy

danielroe (approved):

danielroe:
thank you ❤️

